Issue #118, opened 2011-09-20
Client.prototype.format shouldn't treat "?" as placeholder inside string literals


TEHEK:
## Description

`Client.prototype.format` attempts to replace every occurrence of `?` for a variable. Even if it's inside a string literal.

E.g.:

``` javascript
 mysql.query("SELECT * FROM ponies WHERE emotion = 'wtf?' AND id = ?", [1]) 
```

results in error because function expects two parameters instead of one.
### Option 1

Pretend it's a feature :) And do adhoc stuff like this:

``` javascript
 mysql.query("SELECT * FROM ponies WHERE emotion = 'wtf?' AND id = ?", ['?', 1]) 
```
### Option 2

Placeholders should be left intact if in string literals. Such behavior can be observed in PHP PDO

``` javascript
mysql.query("SELECT * FROM ponies WHERE emotion = 'wtf?' AND id = ?", [1]) 
    // SELECT * FROM ponies WHERE emotion = 'wtf?' AND id = 1
```
### Option 3

Add escaping convention for `?`.

For example, use `#` or `\` and do something like:

``` text
?      => value
#?     => ?
##     => #
###?   => #?
```

IMHO, we should stay away from `\` as it can easily turn into escaping hell like in these non-real-world examples: 

``` javascript
// SELECT * FROM ponies WHERE emotion = 'wtf?' AND id = ?
mysql.query('SELECT * FROM ponies WHERE emotion = \'wtf\\?\' AND id = ?', [1]) 

// SELECT * FROM ninjas WHERE slash_and_question = '\\?' AND id = ?'
// has to be: SELECT * FROM ninjas WHERE slash_and_question = '\\\\\?' AND id = ?'
mysql.query('SELECT * FROM ninjas WHERE slash_and_question = \'\\\\\\\\\\?\' AND id = ?', [1]) 

```

I have patch for options 1 and 2. Option 3 is a piece of cake. Which one should I attach? :)

qwy:
is it still remain open bug?
kinda incredible if it is after 5 year.
i'm using "mysql": "^2.12.0", it's still making trouble there. and yet i cannot find an elegant way out.
either option 2 or 3 makes sense to me.

please do fix it.
or let me know if my info is incorrect.
thanks.

felixge:
@qwy https://github.com/mysqljs/mysql/blob/master/License#L13-L19

PR #122, opened 2011-09-21
Fix for #118 - change #format() to not replace "?" in query string literals

TEHEK:
https://github.com/felixge/node-mysql/issues/118
- Added support for double quotes and backticks
- Added support for proper escaping compliant with mysql specs: 
  - backslash escaping (\" \')
  - escaping with pairing ("" and '' and ``)
- Added support for quote nesting
- Added corresponding test cases (should be enough for most of the cases)
- Wrote some algorithm comments


sondresb:
Is this still unresolved, @felixge? The "simpler way" is not really working, unless exactly the last ? sign is the one you want to keep...

Having "?" as a substitution variable works, but is kind of ugly.

PR #338, opened 2012-11-19
Connection state

### LINKED ISSUE #314: State of mysql connection
Is there any possibilty to check if a connection is currently established?

e.g.
`if ( myConnection.connected )...`

dresende:
Connection states implemented:
- connected
- disconnected
- authenticated
- protocol_error

This should close #314.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -141,7 +141,7 @@ Connection.prototype._handleNetworkError = function(err) {
 };
 
 Connection.prototype._handleProtocolError = function(err) {
-  this.state = err.message || "protocol_error";
+  this.state = "protocol_error";
   this.emit('error', err);
 };
 


Alex-Nabu:
Great to see this feature implemented. Not much documentation on how to use it though. Lets say I'm running a query in a loop. and the query errors out. I want to close the connection and stop whatever I'm doing only I'm stuck in the query callback function so I cant just 'break;' and my only option is to use 'return' and have the query callback not go any further. However The for loop will still continue trying to execute subsequent 'query' function calls and if I on the first query error I had 'destroyed()' or 'end()' the connection. My program will crash telling me the connection has been closed or if I had opted not to close it then the connection is kept open any will get a timeout error later on. How would I use this feature to check the connection state to know if I should break out of my loop or not.

Alex-Nabu:
for(var i= 0; i<offer.items_to_give.length; i++)
								{
									connection.query("DELETE FROM `backpack` WHERE `assetid` = ? AND `appid` = ? AND `contextid` = ?", [offer.items_to_give[i].assetid, offer.items_to_give[i].appid, offer.items_to_give[i].contextid], function(query_error, results, feilds) {
										 if(query_error)
										 {
										 	logger.log("error", "query error : failed to delete items withdrawn from inventory:"+query_error.code, offer);
										 	connection.rollback();
										 	return;
										 }

dougwilson:
Hi @Alex-Nabu , you are probably not going to get a response from a 2 year old issue sitting closed. Please go ahead an open a new issue with your question, please, so everyone can see it. Based on your question, though, the answer is that your issue is because you are not using Node.js async well, as you cannot just call `connection.query` from within a `for` loop like that.

PR #446, opened 2013-04-04
Add bigNumberStrings option for forcing BIGINTs as strings (always)

imkira:
This is in continuation of https://github.com/felixge/node-mysql/pull/437


oliviertassinari:
Side note, this PR doesn't cover the [`OkPacket.insertId`](https://github.com/mysqljs/mysql/blob/e16c52da980e46012a391232be884b797b8af070/lib/protocol/packets/OkPacket.js#L21) value. It's still returning a number. I will do the casting manually.

PR #502, opened 2013-05-28
add `stringifyObjects` option

### LINKED ISSUE #501: Warning: Security Issues with Query-Type-Escaping and Express
Hi, 

if you are using Express and the Express query parser, you might be doing something like `GET /users?id=xxx`

And in your app:

`db.query('SELECT * FROM users WHERE id = ?', [req.query.id])`

I have seen this several times now, this is very dangerous! The Express query parser is translating arrays and objects, which node-mysql loves to translate as well.

So, an attacker could do `GET /users?id[id]=1` which Express translates to `{"id":"1"}` which ends up as

`SELECT * FROM users WHERE id = `id` = '1'`

I understand that this might not be an issue with node-mysql, but it would be great to have an option to turn off the automatic type translation in node-mysql, maybe even by default, since most people are not expecting the escaper to behave like that.

tellnes:
fixes #501

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -140,6 +140,8 @@ When establishing a connection, you can set the following options:
 * `database`: Name of the database to use for this connection (Optional).
 * `charset`: The charset for the connection. (Default: `'UTF8_GENERAL_CI'`)
 * `timezone`: The timezone used to store local dates. (Default: `'local'`)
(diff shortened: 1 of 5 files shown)


tellnes:
https://github.com/mysqljs/mysql/pull/1509

Issue #748, opened 2014-03-05
Support for object translation in WHERE claus


soitgoes:
Currently the following syntax

```
conn.query("select * from user where ?", {id:1,givenName:""}, function(....
```

generates

```
select * from user where `id` = 1, `givenName` = \'\'
```

which isn't valid SQL.

I love this capability in INSERTS, but I would love to see it available in WHERE clauses as well.  Without getting into the full scope of an ORM if we could simply render and as the delimiter then it would handle lots of the string manipulation that one runs into in a project.

```
select * from user where `id` = 1 AND `givenName` = \'\'
```

Thoughts?

dougwilson:
I'm going to close this, as the security community has deemed the features like this large foot guns for security, and I tend to agree. Unfortunately many users never read the documentation and there is an expectation from the security community now that this library should allow any arbitrary user-controlled value in to the placeholder and not effect the query. This means support for your proposal will just not happen within this module. I would suggest looking at a nice, feature-full query builder like https://hiddentao.github.io/squel/v4/index.html for this functionality.

PR #749, opened 2014-03-05
Add support for where claus object parsing

soitgoes:
This addresses issue #748


dougwilson:
So I don't think any agreement was settled on here, and it looks like this PR is changing code that no longer exists in this module. Also, no documention was ever added as request. If still interested, you're welcome to re-make the PR against https://github.com/mysqljs/sqlstring

Issue #833, opened 2014-05-29
Pool leaking connections


seangarner:
I need to code up a minimal test case but I'm finding that the connection pool leaks connections that are released if `changeUser` has been called on it during the lifetime of the connection.

When release is called on the connection it appears that it gets removed from the pool in so much that it's no longer in `pool._allConnections` or `pool._freeConnections`.  But the connection must live on somewhere because the tcp connection stays open.  I'm using the crude `netstat -lanp | grep 3306 | wc -l` to count the connections, and the number grows by the number of connections got from the pool.

If using the connection without calling `changeUser` then it returns the connection to the pool as expected.

PR #837, opened 2014-05-30
release pool connection restores pool user configuration

seangarner:
Discussed potential for this in #833 whereby a connection that is checked out from the pool then has `changeUser` called it would be marked for purging when it was released.  Instead during release `changeUser` is called again to restore it back to the pool configuration before it's released back into the pool.

I've updated the test that covered restoreUser and all tests are passing locally, but I couldn't find a jshint or styleguide so I'm fully open to changes before merging.


seangarner:
superseded by #1088.

/me makes a note to himself to find out how the rebase went so wrong

PR #864, opened 2014-07-01
added connections count getters

quibusus:


dougwilson:
There is now an `acquire` and `release` event on the pool, which would fit what you described and there was never really any response to my last comment. If you still think there is a use-case, please feel free to open a new PR based on the current release, and be sure to include docs & tests which are missing from this PR.

PR #877, opened 2014-07-21
support for POINT class

kevinhikaruevans:
Since node-mysql supports [SELECTing and casting the POINT class](https://github.com/felixge/node-mysql/blob/0c99f5639f4c28ce10fabbac9523b58b61f508d1/lib/protocol/Parser.js#L311), it would make sense to also support escaping `POINT`s during INSERTs and UPDATEs. Currently, if you SELECT a field with a point, it'll return a plain object with `x` and `y` attributes. This PR will allow it to also INSERT and UPDATE an object with `x` and `y` attributes.

I have no idea if this was the correct way to append a new type in `SqlString`, but it works fine for me. Date casts were defined there, so it seemed like a good place to start. The POINT function is defined [here](https://dev.mysql.com/doc/refman/5.0/en/data-types.html).

### DIFF
--- a/test/unit/protocol/test-SqlString.js
+++ b/test/unit/protocol/test-SqlString.js
@@ -117,7 +117,13 @@ test('SqlString.escape', {
 
     assert.strictEqual(string, "'" + expected + "'");
   },
-
+  'points are converted to POINT objects': function() {
+    var expected = 'POINT(123.004, -10.1)';
+    var input    = {x: 123.004, y: -10.1};
+    var string   = SqlString.escape(input);
+    
+    assert.strictEqual(string, expected);
+  },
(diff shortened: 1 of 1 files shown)


lib/protocol/SqlString.js
```diff
@@ -36,6 +36,10 @@ SqlString.escape = function(val, stringifyObjects, timeZone) {
   }
 
   if (typeof val === 'object') {
+    // for the mysql point class
+    if(val.hasOwnProperty('x') && val.hasOwnProperty('y')) {
```
dougwilson: We can also make this a little less risky by saying "if it has x and y _and no other properties_"

leearmstrong:
Any updates on this getting merged in?

Issue #880, opened 2014-07-24
PoolCluster should re-add removed nodes after a period of time


dougwilson:
From https://github.com/felixge/node-mysql/issues/879#issuecomment-49815968 the PoolCluster should have a setting to automatically re-add a removed node after a period of time.

/cc @benthor

PR #906, opened 2014-09-09
PoolСluster auto-restore node

eakorolev:
#880

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ npm-debug.log
 /test/config.js
 
 *.sublime-*
+.idea
 
 .DS_Store
--- a/lib/PoolCluster.js
+++ b/lib/PoolCluster.js
@@ -14,6 +14,7 @@ function PoolCluster(config) {
   config = config || {};
   this._canRetry = typeof config.canRetry === 'undefined' ? true : config.canRetry;
   this._removeNodeErrorCount = config.removeNodeErrorCount || 5;
+  this._restoreNodeTimeout = config.restoreNodeTimeout || 0;
   this._defaultSelector = config.defaultSelector || 'RR';
 
   this._closed = false;
@@ -54,6 +55,7 @@ PoolCluster.prototype.add = function(id, config) {
     this._nodes[id] = {
       id: id,
       errorCount: 0,
+      online: true,
       pool: new Pool({config: new PoolConfig(config)})
     };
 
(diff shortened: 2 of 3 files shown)


eakorolev:
@dougwilson, should I wait for this fix?

jonathandelgado:
+1

dougwilson:
Sorry for the delay :( This PR is marked for the 2.6 milestone and I'm currently working through merging them all.

.gitignore
```diff
@@ -8,5 +8,6 @@ npm-debug.log
 /test/config.js
 
 *.sublime-*
+.idea
```
dougwilson: Please remove this; really, editor-specific entries belong in your global ignore list.

lib/PoolCluster.js
```diff
@@ -124,17 +128,35 @@ PoolCluster.prototype._getNode = function(id) {
...
-
       this._clearFindCaches();
 
-      node.pool.end();
+      if ((this._restoreNodeTimeout > 0) && !node.restoreTimeoutID) {
+        node.online = false;
+        this.emit('offline', node.id);
+
+        node.restoreTimeoutID = setTimeout(function() {
+          delete node.restoreTimeoutID;
```
dougwilson: Set back to 0 or something rather than deleting and changing the hidden class of the node.

lib/PoolCluster.js
```diff
@@ -124,17 +128,35 @@ PoolCluster.prototype._getNode = function(id) {
...
-      delete this._nodes[node.id];
-
       this._clearFindCaches();
 
-      node.pool.end();
+      if ((this._restoreNodeTimeout > 0) && !node.restoreTimeoutID) {
+        node.online = false;
+        this.emit('offline', node.id);
+
+        node.restoreTimeoutID = setTimeout(function() {
```
dougwilson: Do we even need to do this using a timer? Why can we not calculate the restoration time and in functions that get nodes, simply query that? (and if `online` is important, it can be a getter). We may also want to use `process.hrtime` instead of dates and timers such that we are using a monotonically-increasing source that is not subject to system date/time adjustments.

eakorolev: Hi-res time is good idea.
Then, we should create another queue `_offlineNodeIds` (like `_serviceableNodeIds`). Every time we call `_findNodeIds`, it will check first record of `_offlineNodeIds` (id+hrtime) and restore offline node if it is needed. Then restored node should be moved from `_offlineNodeIds` to `_serviceableNodeIds`.

Do you approve this idea?

.gitignore
eakorolev: I will remove this line, and sublime-specific things too :smiley: OK?

lib/PoolCluster.js
eakorolev: You're right. I'll fix it

lib/PoolCluster.js
dougwilson: SGTM

.gitignore
dougwilson: I just went ahead and cleaned up the file on `master` :)

dougwilson:
Please test the version that has landed on `master` as thoroughly as you need so we can catch issues before being released :)

eakorolev:
I've tested pool cluster by shutting down mysql server nodes (I had 3 of them). Works well, but sometimes pool cluster emits a lot of `offline` events at a time. It's not a big problem, because it is not an exception.

dougwilson:
> Works well, but sometimes pool cluster emits a lot of offline events at a time

I'll take a look into this.

eakorolev:
It happens when cluster is trying to restore node, but remote server still offline.

dougwilson:
Gotcha. Yes, I was able to determine the same. It's a limitation unless the `PoolCluster` is seriously refactored such that we make the lookup for a node to be async instead of sync.

Issue #964, opened 2014-12-21
connection error prevents pool from ending connection


afanasy:
This doesn't work as expected (pool doesn't end, script doesn't exit)
```javascript
pool.getConnection(function (err, conn) {
  conn.query('').
    on('end', function () {
      conn.release()
      pool.end()
    })
})
```

however, this works (pool closes connections and script exits)
```javascript
pool.getConnection(function (err, conn) {
  conn.query('').
    on('error', function () {}).
    on('end', function () {
      conn.release()
      pool.end()
    })
})
```

Thanks for the great lib!

dougwilson:
Ah ha, I see now. So what is happening is is a sequence, like query, doesn't have an error handler (either an error event listener or a callback function), then the error is not handed to the sequence, but as a consequence, the end event for the sequence will never fire. Seems like a bug to me, since we're not being consistent on when the end event is emitted. Since it currently emits on an error, it should still emit in this case as well :)

afanasy:
@dougwilson thanks! yes, that would be more expected behaviour.

manuguerra:
Actually the *end* event is being fired; the problem is that, since there are no listeners to the *error* event, it bubbles up to *PoolConnection*, which, in turn, will call *_removeFromPool*. This will remove any references to the connection, and when we try to end the pool, the connection won't be ended/destroyed, causing a stall.
Do we really need to call *_removeFromPool* on *error*, since it will eventually be called on *end* events?

manuguerra:
In my previous pool request, I assumed that all errors would trigger `end` events, but that may not always be the case. All fatal errors trigger `end` events, but it's not necessarily true for other types of errors.
So I thought about making a slight modification of the proposed fix. In `PoolConnection`, we would have:
```javascript
this.on('error', function() {
  self._removeFromPool();
  self.destroy();
});
```
This also fixes the dangling connection issue and passes the tests. 
What do you think, @dougwilson? Is it ok to always destroy the connection here or only if it is in a specific state, like `protocol_error` (which is the case for an invalid query)?

dougwilson:
I believe this issue got fixed at some point, based on me not being able to reproduce anymore. If someone can reproduce with version 2.6.1+, please open a new issue/PR and let me know!

PR #974, opened 2015-01-16
Fix for issue #964: connection error prevents pool from ending connection #964

manuguerra:
When an error event bubbled up to `PoolConnection`, `pool._removeFromPool` was being called, so the connection was being removed from the array but not destroyed. Then, when `pool.end` was called, the connection wasn't being ended, since we didn't have any reference to it anymore.

This pull request adds a test to cover that issue and fixes it by removing the call to `pool._removeFromPool` on connection error, since it's not necessary.

### DIFF
--- a/lib/PoolConnection.js
+++ b/lib/PoolConnection.js
@@ -14,7 +14,12 @@ function PoolConnection(pool, options) {
   // the connection to end as well, thus we only need to watch for the end event
   // and we will be notified of disconnects.
   this.on('end', this._removeFromPool);
-  this.on('error', this._removeFromPool);
+
+  // we need this listener to avoid an exception
+  // but we can't call removeFromPool on error,
+  // otherwise the connection will be removed but not destroyed
+  // which may cause a stall
+  this.on('error', function() {});
 }
 
 PoolConnection.prototype.changeUser = function changeUser(options, callback) {
--- a/test/unit/pool/test-query-error.js
+++ b/test/unit/pool/test-query-error.js
@@ -1,0 +1,40 @@
+var assert = require('assert');
+var common = require('../../common');
+var pool   = common.createPool({
+  connectionLimit : 1,
+  port            : common.fakeServerPort
+});
+var server = common.createFakeServer();
+
+
+server.listen(common.fakeServerPort, function(err){
+  
+  pool.getConnection(function(err, conn) {
+
+    var socket_closed = false;
+
+	// socket should be closed after pool.end()
+    conn._socket.on('close', function() {
+      server.destroy();
+      socket_closed = true;
+    });
+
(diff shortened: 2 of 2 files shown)


dougwilson:
So, adding a function that swallows all errors is not the correct fix; errors need to be handled. The references issue needs to be fixed, yes, but simply swallowing all errors is not the correct way.

manuguerra:
@dougwilson, I agree with you, but is *_removeFromPool* really necessary here, since it's also being called on *end* events?

dougwilson:
Is there a guarantee that all errors will be followed by an end event? Also, how does this fix #964 ? The `end` event is still not being fired on the `query` object.

dougwilson:
> Also, how does this fix #964 ? The end event is still not being fired on the query object.

Nevermind, I read the test incorrectly.

dougwilson:
> but is _removeFromPool really necessary here, since it's also being called on end events? 

Is there a _guarantee_ you can point to me in the code that `end` will always be fired after `error` for where you change? Even then, I still won't add this in; adding an empty function as an event listener is a huge code smell. It tells me there is a bigger problem somewhere and this fixes a symptom of it. We need to fix the actual issue and not add an empty event listener, basically.

dougwilson:
Basically, it sounds like the bug is in the interaction between `_removeFromPool` and `pool.end`.

dougwilson:
If it helps, I looked into this a while ago, and what I found needed to be fixed was https://github.com/felixge/node-mysql/blob/048f2c1040b760f0c15fe800689b590343df0f1c/lib/protocol/Protocol.js#L366-L368 was not correctly getting invoked, leaving the `pool.end` hanging.

dougwilson:
And I believe it was because the error delegation calculation was wrong for the specific case.

dougwilson:
Also, the `on('error', fn)` in the `PoolConnection` may need to call `destroy` on itself, as for what the error listener should probably be doing.

dougwilson:
P.S., I'm not saying your solution here doesn't solve the problem or not, but the change just smells wrong to me, so I'm very iffy in accepting it without knowing more. Ideally there are not empty event listeners, but if there is absolutely no other way to fix it without breaking backwards-compatibility, I'm fine, but essentially I would accept changing it to an empty event listener as a last resort if there is no other possible solution.

manuguerra:
I totally agree with you @dougwilson, and thank you for the clarifications.
I will dig deeper into this issue and will let you know if I have other ideas or questions.

dougwilson:
No problem :) I look forward to this being fixed, and it's really awesome that you included a test! I don't mean to sound harsh or anything :)

manuguerra:
Sure, thanks! :)

PR #982, opened 2015-02-01
Added missing error callback, hanging services

bionicman21:
Added a missing callback that causes any services using the MySQL connection pool to hang when a ping to MySQL fails.  This often happens when connections have been idle for more than 8 hours in a typical MySQL setup, and oftentimes the "ping" request fails to MySQL.

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -88,6 +88,7 @@ Pool.prototype.acquireConnection = function acquireConnection(connection, cb) {
     if (err) {
       pool._connectionQueue.unshift(cb);
       pool._purgeConnection(connection);
+      cb(err);
       return;
     }
 


dougwilson:
A lot of tests are failing now. Also, it's important to note that your change will make the `cb` invoke multiple times now. The reason it wasn't invoked there was because of the line `pool._connectionQueue.unshift(cb)`, which adds the callback to the top of the connection queue and then `pool._purgeConnection` will move the queue forward.

Please adjust your fix to at least not break the existing tests, and ideally add a new test for this, though adding a new test is not a requirement.

bionicman21:
There is one other change that has to happen here.  The noted pull request here will fix the issue where the server hangs on a failed "connection.ping" request.  However, another issue that exists here is that if "connection.ping" fails, the connection is added back to the "_freeConnections" array when the "releaseConnection" function is called from tracing the "_purgeConnection" call.  This will cause an endless loop of frozen connections, exhausting the connection pool.  To fix this, we need to ensure that when the connection.ping fails for an existing connection, we need to remove the connection from the pool and ensure that we do not re-add it to the "_freeConnections" array.

bionicman21:
Ok, I will check on that.  I see your point here.  Please note my above comment as failures in the "connection.ping" request will cause the connection pool to eventually exhaust as it thinks each failed connection is a free connection again.

dougwilson:
@bionicman21 any update on this?

dougwilson:
Closing due to lack of response + incomplete PR.

PR #996, opened 2015-02-16
build: test iojs v1.2.0 on travis-ci

sidorares:


dougwilson:
Update `pacakge.json` to match, please.

sidorares:
You mean "engines" field? I'm trying to understand what's current convention to indicate versions for both node and iojs. Heroku, for example, [suggests](https://devcenter.heroku.com/articles/nodejs-support) using `{ "engines": { "iojs": "1.0.x" } }`. Also all 1.0.x and 1.1.x versions wok just fine. Do we want to test them on travis? If not, should we indicate that they are not supported? ( we don't test against all node patch version anyway)

dougwilson:
> Also all 1.0.x and 1.1.x versions wok just fine. Do we want to test them on travis?

Yes, if we want to support them.

> we don't test against all node patch version anyway

correct, but we do test against all _minor_ node.js versions we support and 1.0 and 1.1 and 1.2 are different _minor_ versions (I'm not suggesting testing against all _patch_ versions, though sometimes we probably should, but eh).

> You mean "engines" field? I'm trying to understand what's current convention to indicate versions for both node and iojs. Heroku, for example, suggests using { "engines": { "iojs": "1.0.x" } }.

Yes, something. Right now, the `engines.node` will make `npm` display a warning to people if they install it on the wrong Node.js (or even just refuse to install if they set that option in their `npm`). I'm not sure what io.js/npm does.

dougwilson:
It's also important to note that I've seen written over and over again on io.js's site that they pledge that io.js is "backward-compatible" with Node.js, where in they say that as long as a library supports Node.js 0.12, then it should be able to automatically run on io.js. This library would fall under that promise, since we won't be using any non-Node.js 0.12 APIs (i.e. nothing io.js-only).

sidorares:
ok, I'll leave this open for now until I investigate behaviour of different npm versions re engines:node/iojs

Even with the promise of that io.js is "backward-compatible" it's good to have extra check that tests do actually run on iojs

dougwilson:
> it's good to have extra check that tests do actually run on iojs

I agree, but it's hard to make a _promise_, at least in the current state (for example, as a Windows user io.js is basically useless because of `node-gyp` issues, etc.). We could certainly test on io.js and add it to the matrix to just allow it to fail, so we'd know if there is something to look into, but it won't _force_ us to look into it.

PR #997, opened 2015-02-16
build: add AppVeyor

dougwilson:
@felixge I need your help here :) Can you please add the necessary webhook? I have sent you an email with the details.


felixge:
@dougwilson webhook should be setup as well.

PR #1001, opened 2015-02-20
Amazon SSL for RDS 2015

winzig:
See http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.Rotation

PR #1004, opened 2015-02-20
added new RDS cert definition

jopann:
See the Amazon announcement about the new RDS certs below.  
https://forums.aws.amazon.com/ann.jspa?annID=2889

The new certs are found at http://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem

This is actually my first github contribution so please let me know if I'm doing it wrong.


jopann:
Dup of #1001

PR #1015, opened 2015-02-25
Update Readme.md

alsotang:
add format example


dougwilson:
The strings are whatever the database sends back, which can be more than just those. Just add a link to the MySQL server documentation, if this is even necessary, rather than trying to enumeriate all possible values.

alsotang:
I think at least we should provide some hints. a hour ago I intend to open this option. In order to get its format, I go into the source code.

dougwilson:
I don't disagree, but the docs here are very misleading and makes it seem like dates come in one format and date time in another. You forgot to add what only time looks like, what those vales will look like when the column definitions have numbers on them, how they vary by MySQL version, etc. Basically the only thing to add to the docs is "they come though as sent by your MySQL server" and that's it. Thu module assumes users know how to use MySQL and would know how their column definitions are setup.

PR #1019, opened 2015-02-27
Optimize SqlString.format

jonapgar:
I was looking for ways to speed up mysql.format after noticing that the performance was *very* slow when attempting to format a query with 100,000+ values. I came up with a couple of improvements, by far the most significant is to use an iterator to step through the values array instead of Array.shift(). See: http://jsperf.com/some-array-reading-comparisons/6

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -66,16 +66,16 @@ SqlString.arrayToList = function(array, timeZone) {
 
 SqlString.format = function(sql, values, stringifyObjects, timeZone) {
   values = values == null ? [] : [].concat(values);
-
+  var index = 0;
   return sql.replace(/\?\??/g, function(match) {
-    if (!values.length) {
+    if (index>=values.length) {
       return match;
     }
 
     if (match == "??") {
-      return SqlString.escapeId(values.shift());
+      return SqlString.escapeId(values[index++]);
     }
-    return SqlString.escape(values.shift(), stringifyObjects, timeZone);
+    return SqlString.escape(values[index++], stringifyObjects, timeZone);
   });
 };
 


jonapgar:
See also:

http://stackoverflow.com/questions/27341352/why-does-a-a-nodejs-array-shift-push-loop-run-1000x-slower-above-array-length-87

jonapgar:
May fix https://github.com/felixge/node-mysql/issues/937

dougwilson:
This sounds good to me, but please make sure that the `values` array stays altered in the same way in the end as it used to be; there are unfortunately people relying on the fact that the array they are passing in gets emptied (or only emptied up to the number of `?` in the SQL).

dougwilson:
Or nevermind my statement above; :) I'll doube-check that for realz and most likely I'll merge your code :O

jonapgar:
I think the values array is cloned anyways in the original code. 

Hold off on that merge. With this change my test suite for my application is failing. Could be an obscure use case. I will update here after I sort it out.

lib/protocol/SqlString.js
```diff
@@ -66,16 +66,16 @@ SqlString.arrayToList = function(array, timeZone) {
 
 SqlString.format = function(sql, values, stringifyObjects, timeZone) {
   values = values == null ? [] : [].concat(values);
```
ronkorving: No need for array copying, as you're not modifying it anymore. More win :)

dougwilson: @ronkorving feel free to make a PR :) And remember that the `values` argument may not always be an array.

ronkorving: Ouch, I missed that it can be an object too. Then there's nothing to gain here. My bad.

PR #1028, opened 2015-03-17
Add support for reset connection packet

jnbhw:
。。。。


dougwilson:
Cannot actually release this until there is a fallback mechanism for server versions that do not support the reset packet, as you can see by the test failures :)

dougwilson:
We also need to figure out a way to actually have the CI severs test this feature as well.

PR #1032, opened 2015-03-20
readme: multiple statement clarification

LinusU:
Allowing multiple statement dosen't in itself expose you to sql injection attacks. It only does so if you already allow unsanitized input in your queries, which you should never do.

Maybe we should even change it to something like:

```diff
-Be careful with this, it could expose you to SQL injection attacks.
+Be careful with this, it could increase the scope of SQL injection attacks.
```

Or just remove the warning, thoughts?


dougwilson:
I like the wording in your comment above compared to what you have in the commit :)

LinusU:
Me too :) I'll update and rebase

LinusU:
Done and done :confetti_ball:

dougwilson:
And merged! :+1:

LinusU:
Thanks! :+1:

Issue #1045, opened 2015-04-01
Timezones integration tests intermittently failing


adrianblynch:
I'm running the integration tests and they fail on:

`node-mysql/test/integration/connection/test-timezones.js:67:12`

With an assertion error:

`"Wed Apr 01 2015 07:50:00 GMT+0100 (BST)" === "Wed Apr 01 2015 08:50:00 GMT+0100 (BST)"`

Throwing in some debugging, the failing case is when the offset is `0`.

I'm running this locally, testing using your guidance from the readme.

`SHOW VARIABLES LIKE "%version%";`

Returns:

|Variable|Value|
|---|---|
|innodb_version|5.6.23-72.1|
|protocol_version|10|
|slave_type_conversions||
|version|5.6.23-72.1|
|version_comment|Homebrew|
|version_compile_machine|x86_64|
|version_compile_os|osx10.9|

Just to add a little more confusion, they passed once with no changes!

Is there any more info I can supply to help debug this?

dougwilson:
Yes, I know they do that on Travis CI. I'm not sure what the reason is. I've asked people in the past to help, but never got a response :( Please, please, if you can figure it out, I would be forever grateful!!

adrianblynch:
I hate dealing with dates but challenge tentatively accepted!

adrianblynch:
I haven't found the underlying cause, but I know what I can do to make the tests pass.

The offsets are popped of the tests array:

`var tests = [0, 1, 5, 12, 26, -1, -5, -20, 'Z', 'local'];`

With '0' being last and failing.

When I reduce down to just that:

`var tests = [0];` 

The tests start to pass.

Adding the others back in, one at a time, the `0` offset test starts to fail again when `local` is introduced back. This means the minimum test to see the failure is:

`var tests = [0, 'local'];`

To sum up, testing `local` first causes `0` to fail.

I'll dig a little bit deeper...

adrianblynch:
I've not looked at this in a while. Still on my todo list though!

dougwilson:
@adrianblynch there no rush :)

dougwilson:
And updates :)?

PR #1062, opened 2015-04-22
Fixed typo in documentation

roccivic:


roccivic:
I'm guessing that the failed build has nothing to do with my commit, but rather an unrelated error in appveyor.

dougwilson:
Failure is from #1045 , don't worry about it.

PR #1088, opened 2015-05-12
switch user back to pool default on connection release

seangarner:
reincarnation of #837...

> Discussed potential for this in #833 whereby a connection that is checked out from the pool then has `changeUser` called it would be marked for purging when it was released.  Instead during release `changeUser` is called again to restore it back to the pool configuration before it's released back into the pool.

> I've updated the test that covered restoreUser and all tests are passing locally, but I couldn't find a jshint or styleguide so I'm fully open to changes before merging.

@dougwilson you said it was ready to merge last year, but I guess something more important came up and I forgot to bump on it.  Would be good to get this change in.

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -109,8 +109,11 @@ Pool.prototype.releaseConnection = function releaseConnection(connection) {
     }
 
     if (connection._purge) {
-      // purge connection from pool
-      this._purgeConnection(connection);
+      // restore the switched user
+      this.restoreUser(connection, function (err) {
+        if (err) return connection.destroy();
+        connection._pool.releaseConnection(connection);
+      });
       return;
     } else if (this._freeConnections.indexOf(connection) !== -1) {
       // connection already in free connection pool
@@ -246,6 +249,29 @@ Pool.prototype._removeConnection = function(connection) {
 
   this.releaseConnection(connection);
 };
+
+Pool.prototype.restoreUser = function restoreUser(connection, callback) {
+  var poolConfig = this.config.connectionConfig;
+  var pool = this;
+
+  var config = connection.config;
+  var user = poolConfig.user;
+  var database = poolConfig.database;
+  var password = poolConfig.password;
+
+  // only when credentials have been changed from the pool defaults
+  if (config.user !== user || config.database !== database || config.password !== password) {
+    connection.changeUser({ user: user, database: database, password: password }, function (err) {
+      if (err) return callback(err);
+      // emit as if it were a new connection so consumers can setup session state again
+      pool.emit('connection', connection);
+      callback(null);
+    });
+    connection._purge = false;
+  } else {
+    process.nextTick(callback);
+  }
+}
 
 Pool.prototype.escape = function(value) {
   return mysql.escape(value, this.config.connectionConfig.stringifyObjects, this.config.connectionConfig.timezone);
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -111,7 +111,7 @@ Pool.prototype.releaseConnection = function releaseConnection(connection) {
     if (connection._purge) {
       // restore the switched user
       this.restoreUser(connection, function (err) {
-        if (err) return connection.destroy();
+        if (err) return this._purgeConnection(connection);
         connection._pool.releaseConnection(connection);
       });
       return;
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -251,17 +251,14 @@ Pool.prototype._removeConnection = function(connection) {
 };
 
 Pool.prototype.restoreUser = function restoreUser(connection, callback) {
-  var poolConfig = this.config.connectionConfig;
   var pool = this;
 
+  var poolConfig = pool.config.connectionConfig;
   var config = connection.config;
-  var user = poolConfig.user;
-  var database = poolConfig.database;
-  var password = poolConfig.password;
 
   // only when credentials have been changed from the pool defaults
-  if (config.user !== user || config.database !== database || config.password !== password) {
-    connection.changeUser({ user: user, database: database, password: password }, function (err) {
+  if (config.user !== poolConfig.user || config.database !== poolConfig.database || config.password !== poolConfig.password || config.charset !== poolConfig.charset) {
+    connection.changeUser(poolConfig, function (err) {
       if (err) return callback(err);
       // emit as if it were a new connection so consumers can setup session state again
       pool.emit('connection', connection);
--- a/lib/PoolConnection.js
+++ b/lib/PoolConnection.js
@@ -8,7 +8,7 @@ inherits(PoolConnection, Connection);
 function PoolConnection(pool, options) {
   Connection.call(this, options);
   this._pool  = pool;
-  this._purge = false
+  this._purge = false;
 
   // When a fatal error occurs the connection's protocol ends, which will cause
   // the connection to end as well, thus we only need to watch for the end event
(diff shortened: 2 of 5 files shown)


seangarner:
Couldn't reproduce the test failure locally with any of those versions of node, but guessing it's a race condition perhaps exposed by windows?  Not sure.  Anyways I've pushed a fix so hopefully the next ones pass.

seangarner:
@dougwilson the test failing now is completely unrelated to these changes, and doesn't happen when I run the tests locally against that version of node.

https://ci.appveyor.com/project/dougwilson/node-mysql/build/107/job/1vquubqcgptk422e#L219
```
test\unit\query\test-streaming-destroy.js
```

I'm going to assume that this is ok to merge unless you say otherwise.

dougwilson:
Ok, so I just pushed a change to `master`. If it fails, then yes, it has nothing to do with your PR. If it doesn't fail, then even though it doesn't seem like it, it would seem very likely something in the PR is causing the failure.

seangarner:
That test didn't fail the first time it ran:

dougwilson:
It's possible your change may have introduced some kind of race condition? I'm not sure, as I haven't taken a hard look at the changes. When running on AppVeyor, there is typically high latency, which really brings out issues in code and tests where you're relying on async events to trigger in specific orders, which is one of the reasons I test on there.

lib/Pool.js
```diff
@@ -247,6 +247,29 @@ Pool.prototype._removeConnection = function(connection) {
...
+  if (config.user !== user || config.database !== database || config.password !== password) {
+    connection.changeUser({ user: user, database: database, password: password }, function (err) {
+      if (err) return callback(err);
+      // emit as if it were a new connection so consumers can setup session state again
+      pool.emit('connection', connection);
+      callback(null);
+    });
+    connection._purge = false;
+  } else {
+    callback();
```
dougwilson: This callback needs to be in a `process.nextTick` so that `restoreUser` is always async instead of sometimes async.

lib/PoolConnection.js
```diff
@@ -29,11 +29,19 @@ PoolConnection.prototype.changeUser = function changeUser(options, callback) {
 
 PoolConnection.prototype.release = function release() {
   var pool = this._pool;
+  var connection = this;
 
   if (!pool || pool._closed) {
     return;
   }
 
+  if (this._purge) {
```
dougwilson: This should be in the pool code, not here; `conn.release()` is just a convenience function, and there are users who actually call `pool.releaseConnection(conn)` instead and they'll end up with different behavior.

dougwilson:
Looks like that test failed on master, so you're changes are in the clear on that :)

lib/Pool.js
seangarner: thanks for spotting that, fixed in 8b93ecb

seangarner:
@dougwilson I've had a look at the travis failure and can't see how this change would have impacted that test.

https://travis-ci.org/felixge/node-mysql/jobs/62480666#L457
```
not ok 144 test/unit/query/test-streaming-destroy.js
  
  events.js:71
          throw arguments[1]; // Unhandled 'error' event
                         ^
  Error: This socket is closed.
      at Socket._write (net.js:519:19)
      at Socket.write (net.js:511:15)
      at FakeConnection.writeRow (/home/travis/build/felixge/node-mysql/test/FakeServer.js:389:18)
      at Timer.exports.setInterval.timer.ontimeout (timers.js:234:14)
```

If you disagree, I'd appreciate any suggestions on where to start looking.

seangarner:
ha... I've just seen the latest master commit.  I'll rebase this branch.

seangarner:
@dougwilson tests are passing and I squashed the review comments ready for merge.

seangarner:
Hi @dougwilson, I don't mean to pester; is there anything else you need me to do before this gets merged in?  It'd be a shame to let it rot out in a PR again.

seangarner:
Anything I can do to help get this merged?

dougwilson:
I'm working on merging this now.

dougwilson:
My biggest debate I've been having with myself is when we should call `changeUser` on the connection: when it is returned to the pool (like now), or when it is requested from the pool (in place of doing a `ping` on that connection).

dougwilson:
Also, would be nice to have a new test specifically testing this, rather then only updating an existing test to account for the change.

test/unit/pool/test-change-user-eject.js
```diff
@@ -38,17 +44,20 @@ server.listen(common.fakeServerPort, function(err) {
 
   pool.getConnection(function(err, conn1) {
     assert.ifError(err);
-    assert.strictEqual(conn1.threadId, 3);
+    assert.ok(conn1.threadId === 1 || conn1.threadId === 2);
+    assert.strictEqual(conn1.config.user, 'user_1');
```
dougwilson: The test needs to run the query `SELECT CURRENT_USER()` to be sure the server-side actually got the change.

lib/Pool.js
```diff
@@ -109,8 +109,11 @@ Pool.prototype.releaseConnection = function releaseConnection(connection) {
     }
 
     if (connection._purge) {
-      // purge connection from pool
-      this._purgeConnection(connection);
+      // restore the switched user
+      this.restoreUser(connection, function (err) {
+        if (err) return connection.destroy();
```
dougwilson: Why call `connection.destroy();` on error instead of `this._purgeConnection(connection);` ?

lib/Pool.js
```diff
@@ -247,6 +250,29 @@ Pool.prototype._removeConnection = function(connection) {
...
+  var poolConfig = this.config.connectionConfig;
+  var pool = this;
+
+  var config = connection.config;
+  var user = poolConfig.user;
+  var database = poolConfig.database;
+  var password = poolConfig.password;
+
+  // only when credentials have been changed from the pool defaults
+  if (config.user !== user || config.database !== database || config.password !== password) {
```
dougwilson: Also needs to verify the charset wasn't changed.

seangarner:
Thanks for the feedback.  Hoping to get time to start working on those changes later.

lib/Pool.js
seangarner: I'm honestly not sure now.  Could have been an arbitary choice.  I've changed to `_purgeConnection` in 590256b and tests pass locally.

seangarner:
> would be nice to have a new test specifically testing this, rather then only updating an existing test to account for the change

The behaviour has changed from purging the connection from the pool, to switching user back to the pool config for the next time it's checked out.  Perhaps it should be renamed?  If not then some guidance on what the 2 separate tests are would be helpful.

test/unit/pool/test-change-user-eject.js
seangarner: Fixed in 8989512d6ccb84e72f8c42590eca258dfd9cadc4

lib/Pool.js
```diff
@@ -247,6 +250,26 @@ Pool.prototype._removeConnection = function(connection) {
...
 };
 
+Pool.prototype.restoreUser = function restoreUser(connection, callback) {
+  var pool = this;
+
+  var poolConfig = pool.config.connectionConfig;
+  var config = connection.config;
+
+  // only when credentials have been changed from the pool defaults
+  if (config.user !== poolConfig.user || config.database !== poolConfig.database || config.password !== poolConfig.password || config.charset !== poolConfig.charset) {
```
dougwilson: there is no such thing as `config.charset`, only `config.charsetNumber`.

Issue #961, opened 2014-12-17
Add Support for Environments that can't handle dynamic require


LorenzGardner:
Some environments may not be able to handle dynamic require statements (nexe a project to bundle up node.js applications into a deploy-able executable).

lib/protocol/packets/index.js & lib/protocol/index.js use "require-all" to add all the .js file in the dir to the module.exports.

This prevents node-mysql from function in places that don't support dynamic require (nexe). Granted there are very few such environments, but anything that attempts to bundle stuff up would be hampered by that (if there are other such things besides nexe).

I'd like to change these two idnex.js file to static list of module.exports.ClientAuthenticationPacket = require('./ClientAuthenticationPacket.js'); ect ..

Pros:
- Can drop the dep on the require-all module
- Support bundling

Cons:
- Adding a new packet/protocol now involves modifying the index.js instead of just adding the .js file

Thoughts?

PR #1134, opened 2015-06-25
allow browserify

t128:
Hi,

you can bundle node packages with browserify. Unfortunately it only works with static requires. So i changed all *require-all* with static requires.

### DIFF
--- a/lib/protocol/packets/index.js
+++ b/lib/protocol/packets/index.js
@@ -1,4 +1,22 @@
-var Elements = module.exports = require('require-all')({
-  dirname : __dirname,
-  filter  : /([A-Z].+)\.js$/,
-});
+module.exports = {
+  ClientAuthenticationPacket: require('./ClientAuthenticationPacket'),
+  ComChangeUserPacket: require('./ComChangeUserPacket'),
+  ComPingPacket: require('./ComPingPacket'),
+  ComQueryPacket: require('./ComQueryPacket'),
+  ComQuitPacket: require('./ComQuitPacket'),
+  ComStatisticsPacket: require('./ComStatisticsPacket'),
+  EmptyPacket: require('./EmptyPacket'),
+  EofPacket: require('./EofPacket'),
+  ErrorPacket: require('./ErrorPacket'),
+  Field: require('./Field'),
+  FieldPacket: require('./FieldPacket'),
+  HandshakeInitializationPacket: require('./HandshakeInitializationPacket'),
+  LocalDataFilePacket: require('./LocalDataFilePacket'),
+  OkPacket: require('./OkPacket'),
+  OldPasswordPacket: require('./OldPasswordPacket'),
+  ResultSetHeaderPacket: require('./ResultSetHeaderPacket'),
+  RowDataPacket: require('./RowDataPacket'),
+  SSLRequestPacket: require('./SSLRequestPacket'),
+  StatisticsPacket: require('./StatisticsPacket'),
+  UseOldPasswordPacket: require('./UseOldPasswordPacket'),
+};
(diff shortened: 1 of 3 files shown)


sidorares:
@t128 what is the point of browserifying node-mysql?

t128:
bundle it into one js file eg: https://github.com/substack/node-browserify/issues/1277#issuecomment-115198436

sidorares:
ok, it might make sense with node-webkit, but why bundle there when you can use native require?

sidorares:
also see related #961

sidorares:
sorry, have not read your comment until the end. ~50mb to ~1mb seems like a lot, but I guess it's not because of one file but because you exclude unused files. This can be done via coverage-like instrumentation to mark files for exclusion before packaging

t128:
yeah, sorry, i changed the comment.

so, it will never happen, as only node.js is really supported?

ps: uglify halved the size to 400kb :)

sidorares:
I'm not completely against your PR, but forcing only static require is actually bad for performance: I often require conditionally features that not always used to make startup time better

dougwilson:
We would only even accept this PR if this library was usable in a browser, but as far as I know, you cannot use it in a browser because you cannot make raw sockets there.

Could you perhaps show/tell us how you are using this module in a web browser?

sidorares:
@dougwilson it's used in node-webkit and bundled for performance considerations, not because of lack of module system

lib/protocol/packets/index.js
```diff
@@ -1,4 +1,23 @@
...
+  ComQueryPacket: require('./ComQueryPacket'),
+  ComQuitPacket: require('./ComQuitPacket'),
+  ComStatisticsPacket: require('./ComStatisticsPacket'),
+  EmptyPacket: require('./EmptyPacket'),
+  EofPacket: require('./EofPacket'),
+  ErrorPacket: require('./ErrorPacket'),
+  Field: require('./Field'),
+  FieldPacket: require('./FieldPacket'),
+  HandshakeInitializationPacket: require('./HandshakeInitializationPacket'),
+  index: require('./index'),
```
dougwilson: Why are you requiring itself? You can remove this.

lib/protocol/sequences/index.js
```diff
@@ -1,4 +1,10 @@
-var Elements = module.exports = require('require-all')({
-  dirname : __dirname,
-  filter  : /([A-Z].+)\.js$/,
-});
+module.exports = {
+  ChangeUser: require('./ChangeUser'),
+  Handshake: require('./Handshake'),
+  index: require('./index'),
```
dougwilson: Why are you requiring itself? You can remove this.

dougwilson:
Oh, gotcha. I left a couple line notes of two things that need to be fixed.

t128:
hmm, you'r right. i missed the *[A-Z]*

dougwilson:
We can probably remove `require-all` from the list of dependencies now, no?

t128:
no, not really: https://github.com/felixge/node-mysql/blob/de5913227dbbaacbbdf22ee38689d090e1451be9/test/run-cov.js#L29

sidorares:
move it to devDependencies ?

t128:
dat error:
```
AssertionError: 'Fri Jun 26 2015 09:50:00 GMT+0000 (Coordinated Universal Time)' === 'Fri Jun 26 2015 09:49:00 GMT+0000 (Coordinated Universal Time)'
```
:(

dougwilson:
@t128 don't worry about that error :) We had a volunteer who would look into it at #1045 but still no resolution.

PR #1142, opened 2015-06-29
Update Readme.md Introduction - no 'database' option in sample code

bl4de:


dougwilson:
The database option is in no way necessary, but I guess it can't hurt to have it in the example.

PR #1150, opened 2015-07-14
Replaced broken link with a more future-proof one

jaredpetersen:
The Holidy Extras "they are hiring" link directed users to a 404 page. I replaced it with the general careers page because the tech jobs section links to a specific job page that will be removed when the position is filled.


jaredpetersen:
Fixes #1151

PR #1180, opened 2015-08-13
Fix streaming bug when using node-mysql with New Relic

### LINKED ISSUE #1153: no response with node 0.12.6
Hi,
I created a pool connection with node-mysql v2.8.0 (same result in v2.5.1).
I use steam and pipe, but I have no error et no response after a few seconds. The process node is still running a few minutes, but no activity.
I tried to downgrade nodejs to v0.10.39 and then it's OK.
I think there is a problem between nodejs 0.12.6 et node-mysql v2.8.0 but I don't know why...
I hope I can help...

kazazes:
Fixes #1153. New Relic wraps function invocations with some voodoo magic, inserting an empty function (with no parameters) along the way. In the Query constructor, node-mysql checks if the `value` parameter is a function and if it is, uses it as a callback. Because a legitimate query callback would expect arguments, checking the `length` property *before* setting `value` as the callback fixes New Relic's intrusion.

This is definitely a fix around New Relic, but because of their pervasiveness (more than one person reported the bug), here's a fix.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -50,7 +50,7 @@ Connection.createQuery = function createQuery(sql, values, callback) {
       options[prop] = sql[prop];
     }
 
-    if (typeof values === 'function') {
+    if (typeof values === 'function' && values.length > 0) {
       cb = bindToCurrentDomain(values);
     } else if (values !== undefined) {
       options.values = values;
@@ -62,7 +62,7 @@ Connection.createQuery = function createQuery(sql, values, callback) {
   options.sql    = sql;
   options.values = values;
 
-  if (typeof values === 'function') {
+  if (typeof values === 'function' && values.length > 0) {
     cb = bindToCurrentDomain(values);
     options.values = undefined;
   }


dougwilson:
Unfortunately this breaks the following 100% legitimate callback function, which is used by a lot of people, which is a very common wrapping pattern:
```js
conn.query('SQL', function () {
  return func.apply(this, arguments)
})

function func(err, results) {
 // ...
}
```

Can you make the fix work with this?

dougwilson:
This patch also seems to break the following use-case:
```js
var domain = require('domain')
var d = d.create()

// ...

conn.query('SELECT 1', domain.intercept(function (rows) {
  // ...
})
```
Because `domain.intercept` returns a new function with no arguments.

kazazes:
Unfortunately there are no other defining characteristics of New Relic's addition. As a workaround, supplying `undefined` as a callback achieves the same effect as my PR, allowing you to stream and use New Relic.

dougwilson:
Gotcha. I mean, unfortunately we cannot actually accept this PR as-is, because it'll break a lot of people :(

PR #1186, opened 2015-08-19
Update RowDataPacket.js

ashwinks:
When using subqueries in select statements, there's no table which means the data would be omitted if using nestTables. This fix only appends data to table objects if there is a table name, otherwise it's appended to the root object.

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -182,7 +182,7 @@ Pool.prototype.end = function (cb) {
 Pool.prototype.query = function (sql, values, cb) {
   var query = Connection.createQuery(sql, values, cb);
   
-  this.emit('query', query);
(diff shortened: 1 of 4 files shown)


dougwilson:
Thank you so much, looks good! Can we get some tests to go along with the change as well :)?

ashwinks:
Sure, will do. Can I add to the existing nestedTable test?

dougwilson:
Adding to the existing tests, as long as it doesn't remove anything, if fine :)

ashwinks:
Doug, I added tests but used the wrong account. Please see https://github.com/ashwinks/node-mysql/commit/c0a85e4ee322c2a9538391c37ed632f7437d36dd

PR #1188, opened 2015-08-20
Update RowDataPacket.js

ashwinks:
When using subqueries in select statements, there's no table which means the data would be omitted if using nestTables. This fix only appends data to table objects if there is a table name, otherwise it's appended to the root object.


dougwilson:
Hi @ashwinks , that's for this PR! I'm looking over it, and it doesn't seem to be backwards-compatible, is that correct? It seems like we were storing the columns under the key `''` before this patch, which it is likely people have probably written code to work-around after these years. I can certainly mark this PR for the 3.0 milestone, though.

PR #1230, opened 2015-09-30
Removed error handler when not having free connections, this error handler  is not possible to be catch

danidomi:
Hi,

In the past I was using 2.1 version from mysql module, but this one was having some issues so I decided to upgrade to 2.9( a lot of good improvements), but when I did that i started to have this issues:

"Connection already released".

Because my code is something like this:

connection.query(sql, queryParams, function(err, rows) {
      connection.release();
.....
           connection.query(sql, queryParams, function(err, rows) {
     			connection.release();
......

I'm not able to bind an error even to the Pool.js

something like this:
pool.on("error", function(err){
    console.log("error");
});

So I'm creating this pull request, so there is a back compactibility.

Thanks :)

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -126,7 +126,7 @@ Pool.prototype.releaseConnection = function releaseConnection(connection) {
     if (this._freeConnections.indexOf(connection) !== -1) {
       // connection already in free connection pool
       // this won't catch all double-release cases
-      throw new Error('Connection already released');
+      // throw new Error('Connection already released');
     } else {
       // add connection to end of free queue
       this._freeConnections.push(connection);


sidorares:
that's your logic error, not runtime's and should bubble up as soon as possible

danidomi:
I aggree, but has I said I was using this in 2.1 and you didn't had this checking (throw new Error), after updating I started having this issue, can you please provided any way for me to bind the error.

AdriVanHoudt:
you should really fetch a new connection from the pool every time

danidomi:
current code:
var poolCluster = mysql.createPoolCluster(clusterConfig)

exports.newConnection = function(callback){
return poolCluster.getConnection('MYSQL*', function(err, connection) {

		if(err){
			callback(err, null);
			return;
		}
		callback(err, connection);

    });

So then I use the newConnection, my question is Im not able to catch that exception by binding an error event? 

Something like pool.on("error", function(err){});

sidorares:
>  I was using this in 2.1 and you didn't had this checking (throw new Error), after updating I started having this issue, can you please provided any way for me to bind the error.

You really need to change your code. It's not that 2.9 is not compatible with 2.1, it's more like 2.1 was not unable to see problem with the way you use pool

danidomi:
That will take too much time for the refactoring + testing.

Sorry for referencing this again, but there is no way for catching that Exception?

AdriVanHoudt:
why would you catch it? If the logic is as in the op it will error every time, this is just doing something logically wrong in your app

dougwilson:
Sorry, that's just not really how minor releases work. We added the new validation because code that triggers it (like your code) can result in really bad behavior where the same connection ends up in the pool multiple times. If you want to take advantage of the new features of this module, you will need to fix your API usage. We have always documented the .release() API such that you are only allowed to call once per connection.

PR #505, opened 2013-06-05
Add Pool options : `maxWait`, `initialSize`, `minIdle`, `maxIdle`

ifsnow:
I did the work improving the function of the `Pool`.
I will be good if these options(`maxWait`, `initialSize`, `minIdle`, `maxIdle`, ..) were helpful.

This is the just beginning. I will improve continuously.

PR #1232, opened 2015-10-05
Extend Pool methods to enable create conns even if there are free ones available

acorbi:
As indicated on https://github.com/felixge/node-mysql/issues/1229 this PR proposes the extension of the functions in the Pool.js module to add a createConnection method that, compared to getConnection, does not acquire a "free" available connection but creates it. createConnection would then be used within getConnection in case there is no "free" available connections.

NOTE: This PR includes only one new unity test related to the new code. Although my plan is to extend the test suite on this PR, I would like to open it now and ask folks over here to provide input on how it could be improved, better tested and in general whether it would be beneficial to integrate this code change in the library. Please provide feedback :+1:

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -28,48 +28,17 @@ Pool.prototype.getConnection = function (cb) {
   }
 
   var connection;
-  var pool = this;
 
   if (this._freeConnections.length > 0) {
     connection = this._freeConnections.shift();
-
     return this.acquireConnection(connection, cb);
   }
 
-  if (this.config.connectionLimit === 0 || this._allConnections.length < this.config.connectionLimit) {
-    connection = new PoolConnection(this, { config: this.config.newConnectionConfig() });
+  return this.createConnection(cb);
 
-    this._acquiringConnections.push(connection);
-    this._allConnections.push(connection);
-
-    return connection.connect({timeout: this.config.acquireTimeout}, function onConnect(err) {
-      spliceConnection(pool._acquiringConnections, connection);
-
-      if (pool._closed) {
-        err = new Error('Pool is closed.');
-      }
-
-      if (err) {
-        pool._purgeConnection(connection);
-        cb(err);
-        return;
-      }
-
-      pool.emit('connection', connection);
-      cb(null, connection);
-    });
-  }
-
-  if (!this.config.waitForConnections) {
-    return process.nextTick(function(){
-      return cb(new Error('No connections available.'));
-    });
-  }
-
-  this._enqueueCallback(cb);
 };
 
-Pool.prototype.getNewConnection = function (cb) {
+Pool.prototype.createConnection = function (cb) {
 
   if (this._closed) {
     return process.nextTick(function(){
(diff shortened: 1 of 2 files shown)


dougwilson:
I'm not sure I understand exactly why this is necessary? If you wanted to "warm" up the pool, the following code works just fine, no?
```js
function warmPool(pool, callback) {
  var connections = [];

  for (var i = 0; i < pool.config.connectionLimit; i++) {
    pool.getConnection(onConnection);
  }

  function onConnection(err, conn) {
    if (!connections && err) return;
    if (!connections) return conn.release();
    if (err) return releaseAll(err);
    connections.push(conn);

    if (connections.length === pool.config.connectionLimit) {
      releaseAll();
    }
  }

  function releaseAll(err) {
    connections.forEach(function (conn) {
      conn.release();
    });
    connections = null;
    callback(err);
  }
}

```

acorbi:
Hi @dougwilson and thanks for your feedback.

Please correct me if I am wrong, but as I understand it, getConnection is going to first check if there are any "free" (created connections that were released and are available on the pool for some thread to acquire them) connections before creating a new one. See https://github.com/felixge/node-mysql/blob/master/lib/Pool.js#L33-L37

Considering this fact, the code you just proposed would try to get **pool.config.connectionLimit** connections. Some of them will be on the list of "free" connections and therefore, the code is not going to create new ones, just acquire and release the "free" ones available previously.

My code change proposes a way of creating those connections ignoring the "free" ones.

Does it make sense?

dougwilson:
I understand the purpose, but can you provide the specific use-case it would serve? Perhaps provide the code you are trying to use this function for?

acorbi:
@dougwilson The use-case is described on https://github.com/felixge/node-mysql/issues/1229 . The goal of the improvement is to reduce the CPU overhead of opening a big amount of connections in little time ( traffic peaks).

Here a code snippet showing how I plan to use it:
```
var connectionDiff = this.config.connectionLimit - this._allConnections.length;
for (var i = 0; i < connectionDiff; i++) {
    pool.createConnection(function(err,conn){
       conn.release();
    });
}
```

dougwilson:
Hi @acorbi , unfortunately your use-case is already covered by the code in https://github.com/felixge/node-mysql/pull/1232#issuecomment-145564634 ; it only adds a larger API surface to this module for no gain over what you can already achieve by just using the existing `getConnection`.

To me, it sounds likely ultimately you want the `initialSize` feature from #505 , which wouold be much easier for people to use. Can you open a new PR porting at least that feature from PR #505 forward to `master` instead?

acorbi:
Hi @dougwilson, I see: On https://github.com/felixge/node-mysql/pull/1232#issuecomment-145564634 you are retaining (not releasing) the connections until all get actually created. Got it!

OK, let me focus on your suggestion of implementing the feature proposed on https://github.com/felixge/node-mysql/pull/505 . It is exactly what I want! 

One question though: Can you kindly shed some light on the status of that PR? It seems to me that it tries to add PoolCluster logic, which is now implemented in the library. So, all the code related to PoolCluster is out-dated now?. I would then extract the logic for the **minIdle**, **maxIdle** and **initialSize** Pool options into a new PR. Sounds good?

dougwilson:
> I would then extract the logic for the minIdle, maxIdle and initialSize Pool options into a new PR. Sounds good?

Sounds perfect.

PR #1233, opened 2015-10-07
Minor performance enhancements in date conversion

danielgindi:


dougwilson:
Seems like an OK change to me.

PR #1302, opened 2015-12-07
Flatten values array

bloodyKnuckles:
If the intent is to flatten the values array this `[].concat(values)` is not doing it for me. Turns out this `[].concat.apply([], values)` does.

For example: http://jsfiddle.net/BloodyKnuckles/d7k7d0Lk/

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -65,7 +65,7 @@ SqlString.arrayToList = function(array, timeZone) {
 };
 
 SqlString.format = function(sql, values, stringifyObjects, timeZone) {
-  values = values == null ? [] : [].concat(values);
+  values = values == null ? [] : [].concat.apply([], values);
 
   var index = 0;
   return sql.replace(/\?\??/g, function(match) {


dougwilson:
Hi! The intent is not to flatten the values. The call is to turn something like a non-array argument (a string for an example) into a one element array. If an array is provided, it will stay as-is.

I hope that explains why we wont be accepting this PR. If you feel like the documentation in the readme can be improved, a PR with documentation change is always welcome!

bloodyKnuckles:
I can dig it. 

Is it okay to do both: convert string values to array elements, AND flatten nested arrays?

`[].concat.apply([], [].concat(values))`

For example: http://jsfiddle.net/BloodyKnuckles/d7k7d0Lk/1/

Or is there a reason NOT to flatten the values array provided to `connection.query`?

dougwilson:
Well, you should be able to see how much just flattening breaks everything by running the tests. For example, the query `field IN (?)` is transformed into `field IN (1,2,3)` when given the argument `[[1,2,3]]`. After your flattening, that would no longer function.

This should be all documented in https://github.com/felixge/node-mysql#escaping-query-values . Let me know if there is a way to make the documentation more clear!

dougwilson:
Basically, no, we will not accept flattening arrays. I was only mentioning the string because you said "If the intent is to flatten the values array this [].concat(values) is not doing it for me.". It is absolutely not our intent to flatten the arguments. You will need to do that yourself if that is your desire. Flattening breaks existing features we provide.

bloodyKnuckles:
I get it (after reading the instructions). I don't need a "?" for every value. Nice!

PR #1331, opened 2016-01-14
Optimize escaping SQL

woollybogger:
In today's V8, appending to a string is faster than building an array and calling `.join()`.

### Benchmark

`$ npm install benchmark`

```js
'use strict';

var Benchmark = require('benchmark');
var SqlString = require('mysql/lib/protocol/SqlString');

var ids = ['id', 'name'];
var arrays = [['value', 12]];
var object = {
  id: 1,
  name: 'Some Name',
  email: 'somename@email.com',
  parentId: 20,
  notify: true,
};

new Benchmark.Suite()
  .add('escapeId', () => {
    SqlString.escapeId(ids);
  })
  .add('escape arrays', () => {
    SqlString.escape(arrays);
  })
  .add('escape object', () => {
    SqlString.escape(object);
  })
  .on('cycle', function(event) {
    console.log(String(event.target));
  })
  .run();
```

#### Results:

Test | Before (ops/sec) | After (ops/sec) | Speedup
------|-----------------------|--------------------|------------
`escapeId` | 1,260,802 | 1,771,151        | 40%
`escape arrays` | 1,037,337 | 2,384,038 | 130%
`escape object` | 279,876 | 456,599        | 63%

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -1,10 +1,12 @@
 var SqlString = exports;
 
-SqlString.escapeId = function (val, forbidQualified) {
+SqlString.escapeId = function escapeId(val, forbidQualified) {
   if (Array.isArray(val)) {
-    return val.map(function(v) {
-      return SqlString.escapeId(v, forbidQualified);
-    }).join(', ');
+    var sql = '';
+    for (var i = 0; i < val.length; i++) {
+      sql += (i ? ', ' : '') + escapeId(val[i], forbidQualified);
+    }
+    return sql;
   }
 
   if (forbidQualified) {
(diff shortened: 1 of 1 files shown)


lib/protocol/SqlString.js
```diff
@@ -1,10 +1,12 @@
...
 
-SqlString.escapeId = function (val, forbidQualified) {
+SqlString.escapeId = function escapeId(val, forbidQualified) {
   if (Array.isArray(val)) {
-    return val.map(function(v) {
-      return SqlString.escapeId(v, forbidQualified);
-    }).join(', ');
+    var sql = '';
+    for (var i = 0; i < val.length; i++) {
+      sql += (i ? ', ' : '') + escapeId(val[i], forbidQualified);
```
dougwilson: The change from `SqlString.escapeId` to just `escapeId` will affect people who are using AOP around this function, since we have been calling it as a method. Just noting this if you wonder why that part is reverted in the merge of your pull request.

lib/protocol/SqlString.js
```diff
@@ -57,11 +59,18 @@ SqlString.escape = function(val, stringifyObjects, timeZone) {
...
-    if (Array.isArray(v)) return '(' + SqlString.arrayToList(v, timeZone) + ')';
-    return SqlString.escape(v, true, timeZone);
-  }).join(', ');
+SqlString.arrayToList = function arrayToList(array, timeZone) {
+  var sql = '';
+  for (var i = 0; i < array.length; i++) {
+    var v = array[i];
+    if (i) sql += ', ';
+    if (Array.isArray(v)) {
+      sql += '(' + arrayToList(v, timeZone) + ')';
```
dougwilson: The change from `SqlString.arrayToList` to just `arrayToList` will affect people who are using AOP around this function, since we have been calling it as a method. Just noting this if you wonder why that part is reverted in the merge of your pull request.

Issue #1349, opened 2016-02-14
"premature close" errors when using query.stream() in conjunction with end-of-stream module


ravenscar:
Using the Query.stream() in conjunction with the reasonably popular https://www.npmjs.com/package/end-of-stream module causes a callback error.

I believe this is because the stream.push() is performed after the 'close' event has been emitted on the stream in Query.js.

PR #1350, opened 2016-02-14
Update Query.js

ravenscar:
#1349 push null value signifying EOF for query items prior to emitting close event on stream.


sidorares:
lgtm

could you add a test @ravenscar ?

ravenscar:
@sidorares I modified the existing unit test which checks for the 'close' event and it now also checks to see that the stream's read buffer is in an ended state. I tested it on the current master and it fails, with this PR it passes.

dougwilson:
Awesome fix!

PR #1354, opened 2016-02-19
Add support for column aliases

nybble73:
I didn't see support for column aliases, so I updated escapeId to see objects in the column list as the signal to alias. The key of the object is the original column and the value the alias.

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -11,6 +11,14 @@ SqlString.escapeId = function escapeId(val, forbidQualified) {
     return sql;
   }
 
+  if( !!val && typeof val === 'object' ) {
+    return Object.keys(val).map(function(orig) {
+      return SqlString.escapeId(orig, forbidQualified) 
+             +  ' AS '
+             + SqlString.escapeId(val[orig], forbidQualified)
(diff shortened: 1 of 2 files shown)


dougwilson:
I'm currently neutral on this pull request. @sidorares what are your thoughts, if any?

sidorares:
not sure I follow. @nybble73 can you give an example query where this would be helpful?

nybble73:
@sidorares - I had, at the time, a use case where I needed to say something like: 

`SELECT code AS title FROM some_table WHERE id=?;`

I actually don't need it anymore, but column aliases are a general thing that I think people need periodically.

adrianblynch:
Maybe I'm missing something, but what about node-mysql stops that query from working at the moment?

Column aliases just work.

Again, maybe I'm missing something.

nybble73:
@adrianblynch This is specifically support within the ?? substitution. ATM there isn't a way to put use aliases there.

dougwilson:
So I don't think any agreement was settled on here, and it looks like this PR is changing code that no longer exists in this module. If still interested, you're welcome to re-make the PR against https://github.com/mysqljs/sqlstring

Issue #1371, opened 2016-03-17
bigint error


k3win:
today, I found an error at bigint parsing.
query result:
```
[ RowDataPacket { trace_id: -185395119379775650, a_timestamp: 1457687553183000 },
  RowDataPacket {
    trace_id: '5083290603631050820',
    a_timestamp: 1457687168117000 },
  RowDataPacket {
    trace_id: '4376063674742127992',
    a_timestamp: 1457686343709000 }]
```
i found that the negative bigint would not been parsed into String and overflowed, and i have set the 'supportBigNumbers' on true
it used to be `-185395119379775634`

dougwilson:
Hi @k3win, thanks for the report! In order to help get this going, can you please provide all the following information?

1. Version of Node.js
2. Version of this module
3. Version of MySQL server
4. DDL I can use to create a sample table
5. SQL I can use to insert sample data into the table
6. Node.js code using this library to select from that table that will demonstrate this issue

Thank you!

dougwilson:
Hi @k3win can you please provide the above information?

k3win:
1.node -v
v5.3.0
2.node-mysql version
"version": "0.4.2"
3.mysql version
mysql  Ver 14.14 Distrib 5.5.44, for debian-linux-gnu (x86_64)
4.ddl
create table test(
	id int(10) primary key AUTO_INCREMENT,
	big bigint
)
5.conf
var config = {
	db:{
		connectionLimit : 10,
		host            : '172.16.30.149',
		user            : 'root',
		password        : 'password',
		database		: 'kdTestDb',
		supportBigNumbers: true
	}
}
6.query
"SELECT * FROM test"

7.results
[{"id":1,"big":"9223372036854775807"},{"id":2,"big":-9223372036854776000},{"id":3,"big":"1111111111111111111"},{"id":4,"big":-1111111111111111200}]

it seems that nodejs do not support the bigint and make it overflow. @dougwilson

dougwilson:
Hi @k3win thanks for the information! Yes, JavaScript does not support them, but we provide some layers to help with that. The main thing I noticed from your report there is you are using _a different module_. The version 0.4.2 corresponds to a different module: https://www.npmjs.com/package/node-mysql

This module is just called "mysql" and the current version is 2.10.2. If you are using that other module, you may want to open a bug on their issue tracker: https://github.com/redblaze/node-mysql/issues

k3win:
[...]

PR #1376, opened 2016-03-29
A bug fix for the Types.LONGLONG number range

Dinwy:
According to MDN(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#Integer_range_for_Number),
Integer range for Number is -9007199254740992 to 9007199254740992.

Bugs In the original code, 
1. There is no range for the negative values.
2. Can't print out 9007199254740993 value. Because of the gt sign at line 103 : Number(numberString) > IEEE_754_BINARY_64_PRECISION.


dougwilson:
Thank you very much! In order to accept this pull request, you need to add tests. Since it is a bug fix, there needs to be at least one test that fails without your change and passes with your change. Please ping me when you add the test to the pull request and I'll take another look :)!

lib/protocol/packets/RowDataPacket.js
```diff
@@ -1,7 +1,6 @@
 var Types                        = require('../constants/types');
 var Charsets                     = require('../constants/charsets');
 var Field                        = require('./Field');
```
dougwilson: Please re-align these statements now that the longest has been removed.

Dinwy:
@dougwilson re-aligned the statements and bug fixed. Please take a look after being tested.

dougwilson:
Thanks, @Dinwy! I still need you to add tests for this. In order to accept this pull request, you need to add tests. Since it is a bug fix, there needs to be at least one test that fails without your change and passes with your change. Please ping me when you add the test to the pull request and I'll take another look :)!

Dinwy:
@dougwilson Oh, I'm sorry. I will add a test soon.

Dinwy:
@dougwilson Could please check this?

ronkorving:
Thanks for the contribution. Does this solve a known bug or did you find it yourself? I wasn't aware of it.

Dinwy:
@ronkorving This will solve a known bug #1371.

test/common.js
```diff
@@ -149,7 +149,8 @@ function mergeTestConfig(config) {
     host     : process.env.MYSQL_HOST,
     port     : process.env.MYSQL_PORT,
     user     : process.env.MYSQL_USER,
-    password : process.env.MYSQL_PASSWORD
+    password : process.env.MYSQL_PASSWORD,
+    supportBigNumbers: true
```
dougwilson: Can you move this to just be set in the one test that needs it, instead of set in every single test? `common.getTestConnection` accepts an optional first argument to provide test-specific configuration to merge with the common config here.

Dinwy: I changed my code and I'm sorry for the additional trouble caused.

test/common.js
```diff
@@ -150,7 +150,7 @@ function mergeTestConfig(config) {
     port     : process.env.MYSQL_PORT,
     user     : process.env.MYSQL_USER,
     password : process.env.MYSQL_PASSWORD
-  }, config);
+    }, config);
```
ronkorving: Needless whitespace change.

Dinwy: How could I miss this. Thanks.

PR #1390, opened 2016-04-17
Improve performance of SqlString.escape()

nwoltman:
Strategies include:
+ Manually escaping strings rather than using a string replace function with a regex
+ Only calling `typeof` on the input value once
+ Not performing extra type checks if the input type has already been determined

#### Benchmark

See [this gist](https://gist.github.com/nwoltman/98dcc294b77da3bedad9bb89e54ab2fa) for the code.

| Value | Before (ops/sec) | This PR (ops/sec) | Speedup (`PR/before`) |
|---------|----------------------|------------------------|-------------|
| `'lastname'` | 4,916,714 | 13,443,318 | 2.73 |
| `'firstname123@gmail.com'` | 4,843,172 | 13,258,714 | 2.74 |
| `'to "escape"'` | 1,846,340 | 3,510,579 | 1.90 |
| `'Lorem ipsum...'` | 1,724,015 | 2,212,050 | 1.28 |
| `'Lorem ipsum...'` with `'\n'` | 845,988 | 1,346,465 | 1.59 |
| JSON | 486,268 | 493,629 | 1.02 |

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -26,40 +26,21 @@ SqlString.escape = function escape(val, stringifyObjects, timeZone) {
   switch (typeof val) {
     case 'boolean': return (val) ? 'true' : 'false';
     case 'number': return val+'';
+    case 'object':
+      if (val instanceof Date) {
+        val = SqlString.dateToString(val, timeZone || 'local');
+      } else if (Array.isArray(val)) {
+        return SqlString.arrayToList(val, timeZone);
+      } else if (Buffer.isBuffer(val)) {
+        return SqlString.bufferToString(val);
+      } else if (stringifyObjects) {
+        val = val.toString();
+      } else {
+        return SqlString.objectToValues(val, timeZone);
+      }
   }
 
-  if (val instanceof Date) {
-    val = SqlString.dateToString(val, timeZone || 'local');
-  }
-
-  if (Buffer.isBuffer(val)) {
-    return SqlString.bufferToString(val);
-  }
-
-  if (Array.isArray(val)) {
-    return SqlString.arrayToList(val, timeZone);
-  }
-
-  if (typeof val === 'object') {
-    if (stringifyObjects) {
-      val = val.toString();
-    } else {
-      return SqlString.objectToValues(val, timeZone);
-    }
-  }
-
-  val = val.replace(/[\0\n\r\b\t\\\'\"\x1a]/g, function(s) {
-    switch(s) {
-      case "\0": return "\\0";
-      case "\n": return "\\n";
-      case "\r": return "\\r";
-      case "\b": return "\\b";
-      case "\t": return "\\t";
-      case "\x1a": return "\\Z";
-      default: return "\\"+s;
-    }
-  });
-  return "'"+val+"'";
+  return escapeString(val);
 };
 
 SqlString.arrayToList = function arrayToList(array, timeZone) {
@@ -155,6 +136,37 @@ SqlString.objectToValues = function objectToValues(object, timeZone) {
 
   return sql;
 };
+
+var charsRegex = /[\0\b\t\n\r\x1a\"\'\\]/g;
+var charsMap = {
+  '\0': '\\0',
+  '\b': '\\b',
+  '\t': '\\t',
+  '\n': '\\n',
+  '\r': '\\r',
+  '\x1a': '\\Z',
+  '"': '\\"',
+  '\'': '\\\'',
+  '\\': '\\\\'
+};
+function escapeString(val) {
+  var chunkIndex = charsRegex.lastIndex = 0;
+  var escapedVal = '';
+  var match;
+
+  while ((match = charsRegex.exec(val))) {
+    escapedVal += val.slice(chunkIndex, match.index) + charsMap[match[0]];
+    chunkIndex = charsRegex.lastIndex;
+  }
+
+  if (chunkIndex === 0) { // Nothing was escaped
+    return "'" + val + "'";
+  }
+  if (chunkIndex < val.length) {
+    return "'" + escapedVal + val.slice(chunkIndex) + "'";
+  }
+  return "'" + escapedVal + "'";
+}
 
 function zeroPad(number, length) {
   number = number.toString();
--- a/test/unit/protocol/test-SqlString.js
+++ b/test/unit/protocol/test-SqlString.js
@@ -76,38 +76,47 @@ test('SqlString.escape', {
 
   '\0 gets escaped': function() {
     assert.equal(SqlString.escape('Sup\0er'), "'Sup\\0er'");
+    assert.equal(SqlString.escape('Super\0'), "'Super\\0'");
   },
 
   '\b gets escaped': function() {
     assert.equal(SqlString.escape('Sup\ber'), "'Sup\\ber'");
+    assert.equal(SqlString.escape('Super\b'), "'Super\\b'");
   },
 
   '\n gets escaped': function() {
     assert.equal(SqlString.escape('Sup\ner'), "'Sup\\ner'");
+    assert.equal(SqlString.escape('Super\n'), "'Super\\n'");
   },
 
   '\r gets escaped': function() {
     assert.equal(SqlString.escape('Sup\rer'), "'Sup\\rer'");
+    assert.equal(SqlString.escape('Super\r'), "'Super\\r'");
   },
 
   '\t gets escaped': function() {
     assert.equal(SqlString.escape('Sup\ter'), "'Sup\\ter'");
+    assert.equal(SqlString.escape('Super\t'), "'Super\\t'");
   },
 
   '\\ gets escaped': function() {
     assert.equal(SqlString.escape('Sup\\er'), "'Sup\\\\er'");
+    assert.equal(SqlString.escape('Super\\'), "'Super\\\\'");
   },
 
   '\u001a (ascii 26) gets replaced with \\Z': function() {
     assert.equal(SqlString.escape('Sup\u001aer'), "'Sup\\Zer'");
+    assert.equal(SqlString.escape('Super\u001a'), "'Super\\Z'");
   },
 
   'single quotes get escaped': function() {
     assert.equal(SqlString.escape('Sup\'er'), "'Sup\\'er'");
+    assert.equal(SqlString.escape('Super\''), "'Super\\''");
   },
 
   'double quotes get escaped': function() {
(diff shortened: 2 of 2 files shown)


dougwilson:
Nice! Code certainly looks weird with the slicing, but I guess if it's faster... :) And thanks a lot for providing that gist with the benchmark code! That would allow us to play with this a bit, for example, to test how it changed very long strings instead of just short ones (only because I have in the past seem optimizations that help one case, but really hurt another case), to have a better overall idea. Otherwise, I think it should be merged soon!

nwoltman:
Thanks! And that's a good idea. I updated the benchmark to include longer strings and it looks like the longer the string gets, the slower my implementation becomes. Would you agree that I should use the original code as a fallback for long strings?

dougwilson:
Perhaps. Are you saying that the longer the string becomes, the more this change actually becomes a de-optimization compared to the original implementation?

I have tried a few times to do "magic numbers" where an implementation will flip to a different one (where one is faster for the short case and one is faster for the long case), but it can sometimes be hard to really know at which number to make the flip (the number may very well different depending on the v8 engine version).

I guess one question I would ask myself is "is this the most optimal implementation, or was it coincidentally optimal for some small case?" What I mean by that is I have seen, when only looking at the small case, that sometimes there are different implementations that can get both short and long cases that were just not immediately obvious when only focusing on the short case.

Anyway, I think that, at the very least, probably adding a magic implementation switch would be good, but also if you want to think some more on the optimization, feel free :)

nwoltman:
Yes, that is what I meant to say.

Unfortunately the only optimizations left that I know of are micro-optimizations, and their small performance benefits are not worth making the code less readable. I know my current implementation cannot be optimized further because of the following: The case where the `.replace()` call is much faster is when the input is a long string that doesn't have any characters to be escaped, so a majority of my code never get's used. In that case, even if I take out all of the unused code, the `.replace()` method is still about 4x faster (against "Lorem Ipsum"). For reference, here's what my code would look like without the unused code:

```js
function escapeString(val) {
  for (var i = 0; i < val.length; i++) {
    switch (val.charCodeAt(i)) {
    }
  }
  return "'" + val + "'";
}
```

So in that sense, it cannot be optimized more. However, this gives us a better idea of why my method is faster when either a) the string is short, or b) the string has many characters that need to be escaped. I'm assuming my method is faster because it is 100% client JS where as with `.replace()` it has to go back and forth between the client code/provided callback and the V8 internals. With the long, plain string, `.replace()` wins because it only goes into the V8 internals once, and the V8 code is faster than just JS code.

Now there are two things that can still be done. The first is trying to find the magic number. The second is putting this somewhere:

```js
if (!/[\0\b\t\n\r\x1a\"\'\\]/.test(val)) {
  return "'" + val + "'";
}
```

Even though that reduces performance a little for short strings and strings that contain characters that need to be escaped, it greatly improves performance for longer strings that do not contain characters that need to be escaped.

One option is to optimize more for the common cases without de-optimizing the rarer cases. Do you have any suggestions on what the common cases might be? From what I've seen, most strings don't have any of those characters that need escaping.

dougwilson:
Awesome, I love this research :) 

> Do you have any suggestions on what the common cases might be?

That is definitely a good question, and I'm not too sure off-hand. Just thinking about some of my MySQL usages that are out there, most of my calls to `.escape` is just with a number primitive, with the second most being strings. From the string perspective, I image the vast majority do not conain any characters that need to be escaped, even long strings. If they do have characters that need to be escaped, it would be due to the single or the double quote characters. Certainly the cases where I am using long strings are typically going to be storing some kind of serialization, either HTML or JSON, both of which typically contain many double quote characters (and single quotes for HTML).

So from just thinking out loud to myself, perhaps a more common case for long strings would be (1) nothing is being escaped at all (perhaps it's some kind of simple CSV?) (2) there are lots of escaped characters within the string. I wonder (2) fares over all.

Also, on the note of optimizations, I noticed that in this module, that giant regular expression is a literal within the function. I feel like I've heard that doing that is not great for v8, and that hoisting the regular expression out and into a global helps a lot. I wonder how much that would help here. The function in the `.replace` could also potentionally be hoisted out since it doesn't close over any variable.

Another thing is that I usually find that regular expressions are very fast to execute in the v8 engine. I wonder if there is a way to replace that `for` loop with a regular expression that finds the next character to escape, rather than looking up the char code for each character? Perhaps that's what's eating up all the time in the new function. I'm thinking something along the lines of:
```js
var chars = /[\0\b\t\n\r\x1a\"\'\\]/g;
function escapeString(val) {
  chars.lastIndex = 0;
  var match;
  while((match = chars.exec(val))) {
    switch (val.charCodeAt(match.index)) {
    }
  }
  return "'" + val + "'";
}
```
That is probably pretty cose to what `.replace` does internally, though not sure if it's faster or not :)

nwoltman:
Based on your suggestions, I created a [new benchmark](https://gist.github.com/nwoltman/9055f5de94e145dddb055134b26a4ab5). The benchmark tests 4 possible solutions:

+ `escapeString0` - The same as the current solution but with the regex and callback hoisted
+ `escapeString1` - Loops over each character and slices + replaces if it sees a character that needs to be escaped (the solution currently in this PR)
+ `escapeString2` - The same as `escapeString1` but returns early if a regex determines that the string has no characters that need to be escaped.
+ `escapeString3` - Uses `regex.exec()` to jump to the parts of the string that need to be escaped and slices + replaces

Here are the results I got with Node v6 (V8 5.0):

---

0: `'lastname'`
```sh
$ node escapeString.js 0
escapeString0 x 5,351,206 ops/sec ±0.64% (90 runs sampled)
escapeString1 x 9,682,148 ops/sec ±0.54% (90 runs sampled)
escapeString2 x 9,878,574 ops/sec ±1.02% (85 runs sampled)
escapeString3 x 10,784,952 ops/sec ±0.56% (90 runs sampled)
Fastest is escapeString3
```

---

1: `'firstname123@gmail.com'`
```sh
$ node escapeString.js 1
escapeString0 x 4,964,859 ops/sec ±0.46% (90 runs sampled)
escapeString1 x 5,930,044 ops/sec ±0.56% (93 runs sampled)
escapeString2 x 11,861,430 ops/sec ±0.64% (89 runs sampled)
escapeString3 x 11,503,335 ops/sec ±0.52% (92 runs sampled)
Fastest is escapeString2
```

---

2: `'line\nbreak'`
```sh
$ node escapeString.js 2
escapeString0 x 2,118,420 ops/sec ±1.23% (88 runs sampled)
escapeString1 x 5,347,677 ops/sec ±0.70% (88 runs sampled)
escapeString2 x 4,134,228 ops/sec ±0.58% (88 runs sampled)
escapeString3 x 3,796,514 ops/sec ±0.43% (93 runs sampled)
Fastest is escapeString1
```

---

3: `'"supercalifra\r\ngilisticexpialidocious"'`
```sh
$ node escapeString.js 3
escapeString0 x 1,348,827 ops/sec ±0.70% (89 runs sampled)
escapeString1 x 2,245,956 ops/sec ±0.46% (91 runs sampled)
escapeString2 x 2,240,128 ops/sec ±0.62% (89 runs sampled)
escapeString3 x 1,576,402 ops/sec ±0.52% (89 runs sampled)
Fastest is escapeString1
```

---

4: `'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.'`
```sh
$ node escapeString.js 4
escapeString0 x 1,576,430 ops/sec ±1.03% (86 runs sampled)
escapeString1 x 365,171 ops/sec ±1.26% (88 runs sampled)
escapeString2 x 2,011,189 ops/sec ±0.53% (90 runs sampled)
escapeString3 x 2,030,396 ops/sec ±0.61% (90 runs sampled)
Fastest is escapeString3
```

---

5: `'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.\nUt enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.\nDuis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.\nExcepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.'`
```sh
$ node escapeString.js 5
escapeString0 x 635,228 ops/sec ±0.57% (89 runs sampled)
escapeString1 x 297,829 ops/sec ±0.63% (84 runs sampled)
escapeString2 x 309,280 ops/sec ±0.50% (89 runs sampled)
escapeString3 x 1,028,440 ops/sec ±0.56% (88 runs sampled)
Fastest is escapeString3
```

---

6: `'{"name":"John Brown","id":123456789,"now":"2016-04-30T02:27:34.096Z","settings":{"on":true,"value":5,"defaultView":"activity"}}'`
```sh
$ node escapeString.js 6
escapeString0 x 372,327 ops/sec ±0.78% (89 runs sampled)
escapeString1 x 466,481 ops/sec ±0.48% (93 runs sampled)
escapeString2 x 467,594 ops/sec ±0.60% (89 runs sampled)
escapeString3 x 330,521 ops/sec ±0.45% (94 runs sampled)
Fastest is escapeString2,escapeString1
```

---

The results are pretty interesting. The performance of each method seems to be a function of the *string length* and the *ratio of escape characters to normal characters*. `escapeString0` and `escapeString3` are slower when there are a lot of characters that need to be escaped and `escapeString1` and `escapeString2` are slower when the string is longer than ~25 characters and there are few characters to escape. `escapeString2` is always fast when there are no characters to escape and `escapeString3` is always fast when there are no or few characters to escape.

I feel like strings 2 and 3 are less common, but it's still helpful to see how each solution performs on them. String 5 is fairly common for things that involve messaging or commenting.

Based on that, I think `escapeString2` and `escapeString3` are the best candidates, although `escapeString3` is the worst with JSON.

nwoltman:
@dougwilson I took your suggestion and used it to create a solution that is faster in all cases (with the help of [this benchmark](https://gist.github.com/nwoltman/9055f5de94e145dddb055134b26a4ab5) - which I had previously made a comment about on this PR but removed because it was outdated and added too much noise here). I updated the code and the benchmark results in the description of this PR.

dougwilson:
@nwoltman awesome!

PR #1410, opened 2016-05-10
Add multiple placeholder example

betweenbrain:
This adds a simple example of using multiple placeholders with an array.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -626,8 +626,8 @@ connection.query('SELECT * FROM users WHERE id = ?', [userId], function(err, res
 Multiple placeholders are mapped to values in the same order as passed. For example:
 
 ```js
-var values = ['a', 'b', 'c']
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -623,6 +623,17 @@ connection.query('SELECT * FROM users WHERE id = ?', [userId], function(err, res
 });
 ```
 
+Multiple placeholders are mapped to values in the same order as passed. For example:
+
+```js
+var values = ['a', 'b', 'c']
+connection.query('INSERT INTO users(foo, bar, baz) VALUES(?, ?, ?)', [userId], function(err, results) {
```
dougwilson: What is the `userId` variable? Did you mean to use the `values` variable here?

betweenbrain: Thanks @dougwilson. That was a complete copy/paste error. I've fixed it and made a slight adjustment for hopefully better clarity.

PR #1430, opened 2016-05-29
Performance improvements for formatting in SqlString (caching)

nwoltman:
Improve performance by:
- Returning early if there are no values to use for formatting
- Not copying the `values` argument if it is already an Array
- Compiling and caching an optimized formatting function for each unique query string
#### Benchmark Results

| Query | Before (ops/sec) | This PR (ops/sec) | Speedup (`PR/before`) |
| --- | --- | --- | --- |
| no values | 1,950,229 | 52,944,361 | 27 |
| 1) simple query | 842,548 | 3,603,032 | 4.3 |
| 2) more values | 700,379 | 2,848,857 | 4.1 |
| 3) a lot of values | 419,132 | 950,663 | 2.3 |
| 4) long query string, few values | 748,445 | 3,538,848 | 4.7 |

See [this gist](https://gist.github.com/nwoltman/7de8311bb69b56bb0002265ad0e5df97) for the code.

The technique for compiling the formatting function was explained by [Felix](https://github.com/felixge) in his talk/post [Faster than C?](https://github.com/felixge/faster-than-c#examples). It's the same technique that [EJS](https://www.npmjs.com/package/ejs) is built on. Basically, this is how it works:

``` js
buildFormatter('SELECT * FROM `user` WHERE ?? = ?');
// returns a function that looks like this:
function formatter(SqlString, values, stringifyObjects, timeZone) {
  "use strict";
  var len = values.length;
  return "SELECT * FROM `user` WHERE " + (0 < len ? SqlString.escapeId(values[0]) : "??") + " = " + (1 < len ? SqlString.escape(values[1], stringifyObjects, timeZone) : "?") + "";
}
```

The `(index < values.length ? escape(values[index]) : 'placeholder')` is needed in case there are less values provided than placeholders.

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -11,6 +11,7 @@ var charsMap   = {
   '\'': '\\\'',
   '\\': '\\\\'
 };
+var formatters = {};
 
 SqlString.escapeId = function escapeId(val, forbidQualified) {
   if (Array.isArray(val)) {
@@ -72,20 +73,20 @@ SqlString.arrayToList = function arrayToList(array, timeZone) {
 };
 
 SqlString.format = function format(sql, values, stringifyObjects, timeZone) {
-  values = values == null ? [] : [].concat(values);
+  if (values == null) {
+    return sql;
+  }
 
-  var index = 0;
-  return sql.replace(/\?\??/g, function(match) {
-    if (index === values.length) {
-      return match;
-    }
+  if (!(values instanceof Array)) {
+    values = [values];
+  }
 
-    var value = values[index++];
+  var formatter = formatters[sql];
+  if (!formatter) {
+    formatter = formatters[sql] = buildFormatter(sql);
+  }
 
-    return match === '??'
-      ? SqlString.escapeId(value)
-      : SqlString.escape(value, stringifyObjects, timeZone);
-  });
+  return formatter(SqlString, values, stringifyObjects, timeZone);
 };
 
 SqlString.dateToString = function dateToString(date, timeZone) {
@@ -148,6 +149,33 @@ SqlString.objectToValues = function objectToValues(object, timeZone) {
 
   return sql;
 };
+
+function buildFormatter(sql) {
+  var valuesIndex = 0;
+
+  function escapingReplacer(match) {
+    if (match.length === 1 && match !== '?') {
+      return charsMap[match];
+    }
+
+    var replacement = '(' + valuesIndex + ' < len ? ' + (
+      match === '?'
+        ? 's.escape(v[' + valuesIndex + '],so,tz)'
+        : 's.escapeId(v[' + valuesIndex + '])'
+    ) + ' : "' + match + '")';
+
+    ++valuesIndex;
+
+    return '" + ' + replacement + ' + "';
+  }
+
+  return new Function(
+    's,v,so,tz',
+    '"use strict";\n' +
+    'var len = v.length;\n' +
+    'return "' + sql.replace(/\?\??|["\r\n]/g, escapingReplacer) + '";'
+  );
+}
 
 function escapeString(val) {
   var chunkIndex = charsRegex.lastIndex = 0;


dougwilson:
> Compiling and caching an optimized formatting function for each unique query string

This is a significant addition and will definitely not land in 2.11.0, for through testing. One thing I notice right away is that it looks like it will increase indefinitely, which is a big problem that will need to be addressed before accepting for sure.

Would it be possible to split this PR apart so we can accept the other changes in the meantime, while the caching is discussed and iterated on?

nwoltman:
I thought you might say something like, so I prepared a backup plan :wink:
See #1431

sidorares:
@nwoltman there was an attempt to factor out some non-core functionality so that others might benefit from optimisations like this one - there is a repo at https://github.com/mysqljs/sqlstring . It's currently very behind master, so as a step one we'll need to sync it. Just heads up, not saying that we should freeze non-factored-out sqlstring right now.

dougwilson:
Funny you mention that module, as I was actually planning to sync it after I got @nwoltman s latest changed merged here :)

dougwilson:
Alrighty, the code has been extracted to the 2.x line of the `sqlstring` module! I would suggest either moving this PR there or refactoring this PR to use that module in this code, whichever you feel is the most appropriate.

nwoltman:
Since it's probably not a good idea to always use a cache that can grow indefinitely, I was thinking maybe [lru-cache](https://www.npmjs.com/package/lru-cache) should be used instead of a plain object and users can configure the cache size limit, let the cache be infinite, or not use the cache at all (I'm thinking this could be the default). Does that sound like the right way to go?

sidorares:
@nwoltman sounds like the right way, I'm doing that in named-placeholders - https://github.com/mysqljs/named-placeholders/blob/89fd88865c39a6d825a0e2b54c5ce14e493c3876/index.js#L75

dougwilson:
@nwoltman, remember that `lru-cache` will not install on Node.js 0.6 or 0.8, so at least for a while, we cannot use that as a dependency in this module.

sidorares:
@dougwilson is that new (lru-cache not compatible with 0.8)? I definitely used it with 0.8 previously, but I dropped 0.8 support some time ago and don't now for sure now

dougwilson:
@sidorares, the dependency definitions are using the `^` symbol, which does not work out-of-the-box on the versions on npm is 0.6, 0.8, and early 0.10 versions.

nwoltman:
Okay, how about having an internal cache class with no limits that just uses a plain object and allowing users to pass in their own cache instance if they want to limit the cache size ([like EJS](https://github.com/mde/ejs#caching))?

dougwilson:
@nwoltman, that could work, as long as it is off by default. And at that point, why not have it so they have to pass in their own cache instance if they ever want it to be on? We should very clearly define what this cache class's contract is, including all methods, how we call those methods (as methods or functions), what happens on throws, how callbacks work etc.

nwoltman:
The benefits of having a simple, internal infinite cache are:

1) It's better for performance and memory.

Formatting `'SELECT * FROM user WHERE ?? = ?'` yields:
```
current x 2,158,570 ops/sec ±0.59% (91 runs sampled)
with simpleObject x 4,031,955 ops/sec ±0.52% (95 runs sampled)
with lru-cache x 2,614,982 ops/sec ±0.67% (95 runs sampled)
```
(although formatting `'UPDATE ?? SET ?? = ?, ?? = ?, ?? = ? WHERE id = ?'` yields:
```
current x 566,961 ops/sec ±1.66% (84 runs sampled)
with simpleObject x 1,066,110 ops/sec ±0.54% (93 runs sampled)
with lru-cache x 919,604 ops/sec ±0.70% (95 runs sampled)
```
)

2) Users who don't have memory concerns won't have to deal with another dependency.

3) The internal cache can be shared among connections/pools, which will reduce memory and garbage collector invocations.

dougwilson:
Yes, but people barely read the documentation as it is. You're just never going to convince me that we will, but default, have an unbounded cache in a core module that is used in servers that are never restarted.

nwoltman:
Sorry for not specifying, I agree it definitely shouldn't be on by default. I was thinking it would be off by default and then:

```js
// Pool/connection config
{formatCache: cacheInstance}, // recommended option
{formatCache: true} // infinite cache
```

dougwilson:
I was just trying to comb through the issues and see we were still talking in this while the `SqlString` was split apart to share with `mysql2` into https://github.com/mysqljs/sqlstring

So besides the conversation above, on the technical side should the caching be in that module, or update the PR to move around how it's integrated into this module?

nwoltman:
It would probably be nice to have the caching in `SqlString` so that `mysql2` could also benefit from it, but I think it makes the most sense to have the caching in this module because `SqlString` is a simple library of escaping functions and shouldn't be concerned with configuration or keeping any sort of state.

nwoltman:
I'm going to close this because the performance/memory trade-off probably isn't worth it and escaping query values is already pretty fast so it's unlikely to be a bottleneck for anyone.

PR #1431, opened 2016-05-29
Performance improvements for formatting in SqlString

nwoltman:
Improve performance by:
+ Returning early if there are no values to use for formatting
+ Not copying the `values` argument if it is already an Array
+ Parsing the SQL string using the technique from #1390

#### Benchmark Results

| Query | Before (ops/sec) | This PR (ops/sec) | Speedup (`PR/before`) |
|----------|----------------------|------------------------|-------------|
| no values | 1,733,030 | 48,704,838 | 28 |
| 1) simple query | 759,563 | 1,925,109 | 2.5 |
| 2) more values | 582,647 | 1,098,441 | 1.9 |
| 3) a lot of values | 366,547 | 507,246 | 1.4 |
| 4) long query string, few values | 647,084 | 1,772,840 | 2.7 |

See [this gist](https://gist.github.com/nwoltman/7de8311bb69b56bb0002265ad0e5df97) for the code.

### DIFF
--- a/lib/protocol/SqlString.js
+++ b/lib/protocol/SqlString.js
@@ -72,20 +72,42 @@ SqlString.arrayToList = function arrayToList(array, timeZone) {
 };
 
 SqlString.format = function format(sql, values, stringifyObjects, timeZone) {
-  values = values == null ? [] : [].concat(values);
+  if (values == null) {
+    return sql;
+  }
 
-  var index = 0;
-  return sql.replace(/\?\??/g, function(match) {
-    if (index === values.length) {
-      return match;
-    }
+  if (!(values instanceof Array || Array.isArray(values))) {
+    values = [values];
+  }
 
-    var value = values[index++];
+  var valuesIndex = 0;
+  var chunkIndex = 0;
+  var placeholdersRegex = /\?\??/g;
+  var result = '';
+  var match;
 
-    return match === '??'
-      ? SqlString.escapeId(value)
-      : SqlString.escape(value, stringifyObjects, timeZone);
-  });
+  while (valuesIndex < values.length && (match = placeholdersRegex.exec(sql))) {
+    var value = values[valuesIndex++];
+
+    result += sql.slice(chunkIndex, match.index) + (
+      match[0] === '??'
+        ? SqlString.escapeId(value)
+        : SqlString.escape(value, stringifyObjects, timeZone)
+    );
+
+    chunkIndex = placeholdersRegex.lastIndex;
+  }
+
+  if (chunkIndex === 0) {
+    // Nothing was replaced
+    return sql;
+  }
+
+  if (chunkIndex < sql.length) {
+    result += sql.slice(chunkIndex);
+  }
+
+  return result;
 };
 
 SqlString.dateToString = function dateToString(date, timeZone) {
--- a/test/unit/protocol/test-SqlString.js
+++ b/test/unit/protocol/test-SqlString.js
@@ -180,5 +180,15 @@ test('SqlString.format', {
 
     var sql = SqlString.format('?', { toString: function () { return 'hello'; } }, true);
     assert.equal(sql, "'hello'");
+  },
+
+  'sql is untouched if no values are provided': function () {
+    var sql = SqlString.format('SELECT ??');
+    assert.equal(sql, 'SELECT ??');
+  },
+
+  'sql is untouched if values are provided but there are no placeholders': function () {
+    var sql = SqlString.format('SELECT COUNT(*) FROM table', ['a', 'b']);
+    assert.equal(sql, 'SELECT COUNT(*) FROM table');
   }
 });


lib/protocol/SqlString.js
```diff
@@ -72,20 +73,41 @@ SqlString.arrayToList = function arrayToList(array, timeZone) {
...
+  if (values == null) {
+    return sql;
+  }
 
-  var index = 0;
-  return sql.replace(/\?\??/g, function(match) {
-    if (index === values.length) {
-      return match;
-    }
+  if (!(values instanceof Array)) {
```
dougwilson: Can this be made `Array.isArray(values)`? Is that measurably slower at all?

lib/protocol/SqlString.js
```diff
@@ -72,20 +73,41 @@ SqlString.arrayToList = function arrayToList(array, timeZone) {
...
+    values = [values];
+  }
+
+  var valuesIndex = 0;
+  var chunkIndex = placeholdersRegex.lastIndex = 0;
```
dougwilson: So after looking this over and playing around, I realize hat we need to make sure that `SqlString.format` is re-entrant, since we are calling out to non-private functions (they are properties on the `SqlString` object that can technically be overwritten) within the `while` loop. How much of a performance hit would it be to construct the `placeholdersRegex` on each `.format()` call?

dougwilson:
Thanks, again, @nwoltman! I added you to the `package.json` over the weekend as well :) LGTM besides the two comments I have.

lib/protocol/SqlString.js
nwoltman: `instaceof` is twice as fast in V8 5.0 and historically has been at least 1.5x faster.
```
Array.isArray - true x 52,454,755 ops/sec ±0.90% (90 runs sampled)
instanceof Array - true x 103,139,337 ops/sec ±1.47% (93 runs sampled)
Array.isArray - false x 50,425,812 ops/sec ±0.80% (91 runs sampled)
instanceof Array - false x 97,070,525 ops/sec ±1.42% (86 runs sampled)
```

The only array `Array.isArray` will catch that `instanceof` won't is an array from a `vm`, but I'd assume most people pulling the Array constructor from a `vm` will be following [this snippet](https://gist.github.com/CrabDude/1100841) which lets `instanceof` work correctly.

dougwilson: Gotcha. I would just say that we have to make sure the behavior from the old stays the same after this change. For example, `[1,2,3]` ends up very different in the SQL than `[[1,2,3]]`, so we don't wan to accidentally start not flattening some of the arguments with the change.

lib/protocol/SqlString.js
nwoltman: The performance hit won't be very big, definitely worth making the function re-entrant :)

lib/protocol/SqlString.js
nwoltman: Well I'd hope unit tests would enforce that the behavior stays the same, but here's a very informal proof that it will stay the same with these changes:

`values` can be one of:

1) a non-null value
```js
// values = 100
[].concat(values) // -> [100]
[values] // -> [100]
```

2) an array
```js
// values = [1, 2]
[].concat(values) // -> [1, 2]
// In this PR values would be unchanged, so it would remain [1, 2]
```

The only reason `[].concat()` was being used before was because the function previously modified the `values` array, so a copy was needed to not modify the user's array, but that was changed in f39856c8b1edeb275a0daaccafe6b62f35e971a8 so the copy is no longer necessary.

dougwilson: > Well I'd hope unit tests would enforce that the behavior stays the same

If I had written the module from the ground up, it probably would :) But we have what we have until it's improved upon :)

So, specifically, what I'm asking is that now we are not doing `values = [].concat(values)` and instead doing `values = values instanceof Array ? values = [values]`. Does the `instanceof` check exactly match what `.concat()` is doing interally? You brought up the different from the `vm` (arrays from a different context). We we get one of those arrays, where `instanceof Array` is `false`, do we still behave the same as the `.concat` pattern did?

dougwilson: Also, just thought of this: if `Array.isArray` is much slower than `instanceof Array`, could we get the best of both cases with `value instanceof Array || Array.isArray(value)`?

nwoltman: It looks like `.concat()` uses `Array.isArray()` internally, so no the behaviour would not be the same.

> could we get the best of both cases with `value instanceof Array || Array.isArray(value)`?

Yes, it won't be as fast as just using `instanceof`, but it will still be faster :)

```sh
# values = 'string'
values instanceof Array x 71,716,654 ops/sec ±1.51% (86 runs sampled)
values instanceof Array || Array.isArray(values) x 39,596,842 ops/sec ±0.94% (89 runs sampled)
[].concat(values) x 2,426,460 ops/sec ±0.76% (89 runs sampled)
```

nwoltman:
Updated, waiting on AppVeyor (@dougwilson do you have the power to cancel those intermediary commits I pushed that got queue up for testing on AppVeyor?)

dougwilson:
> @dougwilson do you have the power to cancel those intermediary commits I pushed that got queue up for testing on AppVeyor?

Yep! Canceled them.

lib/protocol/SqlString.js
dougwilson: > Yes, it won't be as fast as just using  instanceof , but it will still be faster :)

Yea, I figured that the non-array case would get extra slow from two checks. I wonder if throwing a `typeof values === 'object'` somewhere will help in that case. Perhaps a `typeof values === 'object' && (values instanceof Array || Array.isArray(values))`?

nwoltman: I don't think it's worth it to make the code that complicated when we just checking if it's an array (but alas, such is JavaScript). Plus, that would be adding a third layer of slowness if `values` is a plain object.

PR #1438, opened 2016-06-07
test(tests): Add more things to test in LOAD DATA INFILE tests

evansiroky:
Adding in more columns of varying types and tests to check validity of inserted data.

### DIFF
--- a/test/fixtures/data.csv
+++ b/test/fixtures/data.csv
@@ -1,4 +1,4 @@
-1,Hello World
-2,This is a test
-3,For loading data from a file
-4,中文内容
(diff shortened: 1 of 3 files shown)


dougwilson:
Thanks for the report. I have tracked this down to a regression in the packet writing with the new pool implementation. Will have this fixed in a jiffy!

dougwilson:
@evansiroky, thanks again for reporting this! It is fixed in 2.11.1, which has been published to npm!

evansiroky:
Thanks for providing a fix so quickly!

dougwilson:
it's no problem at all!

PR #1456, opened 2016-06-29
Fix transaction sample code avoid connection errors

ryutamaki:
I got this error if I don't put `connection.end()` after commit is success

```
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Connection lost: The server closed the connection.
    at Protocol.end (/path/to/node_modules/mysql/lib/protocol/Protocol.js:109:13)
    at Socket.<anonymous> (/path/to/node_modules/mysql/lib/Connection.js:115:28)
    at emitNone (events.js:91:20)
    at Socket.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:926:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1010,6 +1010,7 @@ connection.beginTransaction(function(err) {
           });
         }
         console.log('success!');
+        connection.end();
       });
     });
   });


dougwilson:
Thanks, but the example doesn't include the `createConnection`, either. The example assumes you are making these calls as part of a longer-lived connection.

PR #1462, opened 2016-07-01
Add sql to error messages.

scottaj:
We've been using this change on our project and it has greatly improved our debugging. We're using it in Loopback and we would often get errors straight from the mysql package, but not have any idea which query generated them.

The change is a little hacked in as is, but I'm open to modifying it as needed to get merged.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1074,6 +1074,10 @@ object. Additionally they typically come with two extra properties:
 [Error]: https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Error
 [MySQL server error]: http://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html
 
+In addition, MySQL server errors will have an `err.sql` property that contains the SQL
+of the failed query. This can we useful when using a higher level interface like an
+ORM that is generating your queries for you.
+
 Fatal errors are propagated to *all* pending callbacks. In the example below, a
 fatal error is triggered by trying to connect to an invalid port. Therefore the
 error object is propagated to both pending callbacks:
--- a/lib/protocol/sequences/Sequence.js
+++ b/lib/protocol/sequences/Sequence.js
@@ -52,6 +52,7 @@ Sequence.prototype._packetToError = function(packet) {
   err.code = code;
   err.errno = packet.errno;
   err.sqlState = packet.sqlState;
+  err.sql = this.sql;
 
   return err;
 };
--- a/test/unit/query/test-error-event.js
+++ b/test/unit/query/test-error-event.js
@@ -7,11 +7,13 @@ var server = common.createFakeServer();
 server.listen(common.fakeServerPort, function (err) {
   assert.ifError(err);
 
-  var query = connection.query('INVALID SQL');
+  var sql = 'INVALID SQL';
+  var query = connection.query(sql);
 
   query.on('error', function (err) {
     assert.ok(err, 'got error');
     assert.equal(err.code, 'ER_PARSE_ERROR');
+    assert.equal(err.sql, sql);
     assert.ok(!err.fatal);
     connection.destroy();
     server.destroy();


dougwilson:
Hi @scottaj, I think adding the query makes sense (not *not* in the actual message). Can you follow this up with the additional work to get this into a state ready for us to accept?

1. Make sure no existing tests fail.
2. Add tests that cover your new feature.
3. Add documentation for your new feature.

Thank you!

scottaj:
> I think adding the query makes sense (not not in the actual message)

Just to be clear, you think adding the query is okay, but not in the actual error message?

I'll work on the tests ASAP, I just wanted to get the change reviewed first.

scottaj:
Just wanted to post this as a point in favor of having the query in the error message. This is from a run of the project test suite, and the addition of the query text make the otherwise somewhat ambiguous error really obvious.

image

dougwilson:
Hi @scottaj, yes, I meant to only put it in the `.sql` property and not the message. I absolutely disagree it helps in any way with your example, and it is going to lead to a lot of issues being in the message, as a query can be many kilobytes in size, and it is going to really mess up people's logging if a dozen KB string appears in there.

scottaj:
Fair enough, it's your call. Do you think there is any value into looking at adding it to the message under some type of config option, maybe in another PR? FWIW, keeping it out of the message means that my team will probably continue to use our own fork with it in there because for us it is a night and day difference as far as debugging goes.

I'll go ahead and move forward with pulling it out of the message and taking care of testing and documentation.

Speaking of testing, the CI build failed on about half of the node versions it ran on on a specific test with a timeout error. I can't seem to reproduce it locally. Is that a know issue with the CI suite, or is it likely a regression I added?

dougwilson:
> Do you think there is any value into looking at adding it to the message under some type of config option, maybe in another PR?

No, because that's not a good use of the `Error` object. I understand that you are probably using very simplistic error logging that only logs the error's `.message` property, and you may want to revisit that decision at some point. I can't image that you cannot add into your logging solution a `if` statement to append the `.sql` property to your logged message when logging an error object, at the very least.

> Speaking of testing, the CI build failed on about half of the node versions it ran on on a specific test with a timeout error. I can't seem to reproduce it locally. Is that a know issue with the CI suite, or is it likely a regression I added?

I'm not sure, as I've never had that problem. You did alter the error message and we check that all over the tests, so are you sure that that doesn't have something to do with it? The `master` branch ran clean just 9 days ago when it was last run. If you can investigate why the failures are happening if you don't think it was your test, that would be useful as well :) It's possible something changed in Travis CI between now that then (and the issue would prevent me from merging any PR until I can get it addressed, anyway).

ZJONSSON:
I absolutely agree with @dougwilson that the sql should not be in the message.

Thinking out loud - There are some security concerns to be mindful of.  The sql query itself can easily contain sensitive information (ssns, table structure, internal ids) that should not be exposed to someone who might receive the sql error (either the end client by accident, third party log analyzers etc).  

One way to prevent this kind of a leak is to define the `.sql` property as `enumerable: false` - in which case the `.sql` does not show anywhere unless you specifically ask for it.

```js
Object.defineProperty(err,'sql',{value: sql, enumerable: false});
```

In this the `.sql` wont show up through stringify or looping through keys.  The only way to look at the sql is by accessing it directly, i.e. `err.sql` (which you can easily do in a custom error logger)

scottaj:
I updated the code to remove the query from the error message. I also added a unit test and some documentation. 

My CI build is still failing, and it looks like the other two most recent PRs are failing with the same issue, so I am pretty confident it is not a regression. It looks like it is some kind of race condition in `test-connect-timeout`. I tried to debug locally, but I could not reproduce it and I am not familiar enough with the code to trace down where it might be coming from easily. If someone could point me towards where I might start looking for the issue or tips on how I might reproduce it locally, I'd be happy to debug more.

dougwilson:
> My CI build is still failing, and it looks like the other two most recent PRs are failing with the same issue, so I am pretty confident it is not a regression. It looks like it is some kind of race condition in test-connect-timeout. I tried to debug locally, but I could not reproduce it and I am not familiar enough with the code to trace down where it might be coming from easily. If someone could point me towards where I might start looking for the issue or tips on how I might reproduce it locally, I'd be happy to debug more.

I think I have the issue fixed now, though I don't think it will reflect in your PR unless you rebase the changes on top of the current master.

Readme.md
```diff
@@ -1074,6 +1074,10 @@ object. Additionally they typically come with two extra properties:
 [Error]: https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Error
 [MySQL server error]: http://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html
 
+In addition, MySQL server errors will have an `err.sql` property that contains the SQL
```
dougwilson: Can you move this up to the list right above that enumerates all the properties? Seems weird for this to randomly be it's own paragraph when there is an enumeration of the properties just above.

lib/protocol/sequences/Sequence.js
```diff
@@ -52,6 +52,7 @@ Sequence.prototype._packetToError = function(packet) {
   err.code = code;
   err.errno = packet.errno;
   err.sqlState = packet.sqlState;
+  err.sql = this.sql;
```
dougwilson: I think @ZJONSSON has a good point on the enumerability, if you want to change it.

test/unit/query/test-error-event.js
```diff
@@ -7,11 +7,13 @@ var server = common.createFakeServer();
 server.listen(common.fakeServerPort, function (err) {
   assert.ifError(err);
 
-  var query = connection.query('INVALID SQL');
+  var sql = 'INVALID SQL';
```
dougwilson: If you can un-DRY the test and just put the literal SQL below as well, that would be good :) I don't like tests that are too DRY because it cases accidents.

lib/protocol/sequences/Sequence.js
dougwilson: And, because `this.sql` is not defined in the constructor of `Sequence`, I think the adding of this property should be done by the object that is handing that property and wrapping this super method.

magicdawn:
No matter what way~ pls land this feature in the module 😂

magicdawn:
How can I move along this?

test/unit/query/test-error-event.js
magicdawn: @dougwilson  what is `DRY` ?

dougwilson: Hi @magicdawn sorry I didn't provide more content to my comment above :) So "DRY" is an acronym, but essentially my comment is just saying to remove the `sql` variable and just use the string `'INVALID SQL'` in the file twice.

Issue #605, opened 2013-10-04
DATE type casting


MattMur:
DATE type casting should be turned off due to incorrect formatting it receives when it gets converted back a string. It should return from MYSQL as YYYY-MM-DD, however after type casting is shows as: yyyy-MM-ddTHH:mm:ss.000Z. If you store a field as DATE, why would you want an arbitrary time appended?

SystemParadox:
@dougwilson `dateStrings: true` is all very well, but that also affects TIMESTAMP, which does not seem desirable.

I agree with @CraigLRussell. I would expect the following behaviour:

- TIME and DATE should always return strings - these types should never, **ever** be affected by timezone. JavaScript's Date object just can't be trusted with them. This should return strings even with `dateStrings: false`.
- TIMESTAMP represents a specific instant in time (i.e. unix timestamp relative to UTC) and thus requires whatever timezone handling is necessary to allow Date objects both in and out. Possibly this should still return Date objects even with `dateStrings: true` (possible a separate `timestampStrings` option would be best).
- DATETIME - not sure about this one. Sometimes these are used like TIMESTAMP, sometimes like DATE+TIME. This is what `dateStrings` should specify.

The current behaviour really needs to change. DATE and TIME are usable with `dateStrings: false`, and yet this is the default?

SystemParadox:
@dougwilson any thoughts on this?

I have run into problems with this so many times it's not funny. DATE fields absolutely **MUST NOT** use a timezone-aware format. It is far too fragile.

If the client/server use different timezones, one creates a date object with a time of midnight, sends it over the wire, it gets converted to the other timezone - if you're lucky it'll be 01:00, if you're unlucky it'll be 23:00 and now your date is wrong.

Worse, this will also break if the host environments use a differing timezone libraries, which is just horrific. Try 1970-06-28 with your timezone set to Europe/London on Linux and Windows and you will get different results.

I don't want to set `dateStrings: true` because that loses the timezone information from DATETIME and TIMESTAMP which need it. The workaround is to use a custom typeCast function:

```
var db = mysql.createConnection({
    typeCast: function (field, next) {
        if (field.type == 'DATE') {
            return field.string();
        }
        return next();
    }
})
```

dougwilson:
We can do this in 3.0, which does not have any target yet. It's the reason I haven't closed the issue :)

ravi:
[...]

PR #1481, opened 2016-07-21
 Provide new dateStrings option: "DATE"

bbito:
Adding checks to allow true OR "DATE" as dateStrings values "DATE" should only effect DATE fields whereas true retains original behavior, effecting DATE, DATETIME and TIMESTAMP fields. See: #605 , strongloop/loopback-connector-mysql#120 , strongloop/loopback-connector-mysql#149


bbito:
My apologies, I didn't mean to leave that big comment block in there! I'll clean that up in 12 hours.

bbito:
This hopes to be a low-impact method to satisfy users wishing to handle only DATE fields as strings while retaining the default behavior for TIMESTAMP and DATETIME fields and providing backwards compatibility with existing code. Another and perhaps more efficient approach would be to add yet another property to connection.config with boolean options, e.g. connection.config.dateOnlyStrings.

bbito:
Can anyone provide hints as to why the Travis CI is failing for Node.js: 0.12? I have fixed all other failures, but I have no clue why test-connect-timeout.js would be effected by this change:
`not ok 46 test/unit/connection/test-connect-timeout.js

  

  assert.js:86

    throw new assert.AssertionError({

          ^

  AssertionError: 'ENETUNREACH' == 'ETIMEDOUT'

      at process.<anonymous> (/home/travis/build/mysqljs/mysql/test/unit/connection/test-connect-timeout.js:29:10)

      at process.emit (events.js:129:20)`

-Thanks in advance for any insights!

lib/protocol/packets/RowDataPacket.js
```diff
@@ -61,7 +61,7 @@ function typeCast(field, parser, timeZone, supportBigNumbers, bigNumberStrings,
     case Types.DATETIME2:
     case Types.NEWDATE:
       var dateString = parser.parseLengthCodedString();
-      if (dateStrings) {
+      if (dateStrings === true || field.type === Types.DATE && dateStrings.toString().toLowerCase() === "date") {
```
dougwilson: I'm not sure if this is backwards-compatible, because code today with `dateStrings: 1` will no longer work after this change, since we allowed any truthy value before to mean date strings.

bbito: Thanks for the feedback, I'll look at supporting any truthy value - except `dateStrings.toString().toLowerCase() === "date"` - for backward compatible behavior and also look to adding the requested tests.
I may be back to ask for help on setting up the test.

dougwilson: Sounds good! You may want to rebase your changes to the current master as well, as I just saw the GitHub is saying the PR is not mergable due to conflicts.

bbito: @dougwilson , I was hoping to find an existing test for the current `dateStrings` connection option to either add to or use as a base, but I'm not finding it. Does such a test exist?

dougwilson: I'm not sure.

bbito:
I feel like I've got a good PR now and an additional improvement for ths, more backwards compatible code is that as in the pre-feature code, there is a simple evaluation for truthiness of dateStrings and with this PR, now the additional checks only affect projects with a truthy setting for dateStrings. My earlier attempt would impact performance of both truthy and non-truthy or unset dateStrings configurations.
- So thanks for prodding for that change!
Now for testing, @dougwilson - is there an example you might point me to that performs a similar test for different connection options?

dougwilson:
I can't tell what is in the PR anymore, as it seems like the GitHub changes tab is not working and showing a while lot of other changes: https://github.com/mysqljs/mysql/pull/1481/files

bbito:
Wow, I wouldn't doubt that I messed up the rebase (my first one!), but I think I would have noticed it was now showing 33 files changed instead of 2! I'll see if I can figure out what went haywire!

dougwilson:
> is there an example you might point me to that performs a similar test for different connection options?

All you need to do is create the connection with option you are trying to test in the test. THe current main test that is supposed to be testing all the type casting from the database to JavaScript is https://github.com/mysqljs/mysql/blob/c19c867d032a89263dc8717a84596275b3e4dfa5/test/integration/connection/test-type-casting.js

bbito:
Thanks @dougwilson , I'll look a that existing test.
I'm a bit stumped about the status of the PR... If I look at my repo it looks as I'd expect with 13 commits and 2 files changed: https://github.com/mysqljs/mysql/compare/master...bbito:master

dougwilson:
Weird. I have a comment I want to make, but not sure if it'll get lost or not. Can you do like a squash rebase on top of the current master, perhaps? The following commands may get you out of the hole:
```
$ git fetch origin
$ git reset --hard origin/master
$ git checkout HEAD@{1} -- .
$ git commit -a
# validate the current commit contains all your changes
$ git push -f
```

bbito:
Thanks for the git help - I'll try that now

dougwilson:
Sorry, I forgot you're a fork :) The first two should be the name of your remote instead of origin. By default, this is usually upstream:
```
$ git fetch upstream
$ git reset --hard upstream/master
```

bbito:
Okay, commits squashed down to 1 and back to 2 files changed as hoped - thanks.
Coding is in addition to full-time+ job, so it may take a day or two to get the test going.

bbito:
Test added, please let me know if there are any additional tasks to sew up this PR.

bbito:
Hi @dougwilson , I have a new test that I haven't pushed yet (test-datestrings-options.js), that loops through various truthy, falsy and 'date'-ish values for `dateStrings:`. Should I delete test-datestrings-option-date.js and add test-datestrings-options.js?

bbito:
@dougwilson I stuck the test-datestrings-options.js test onto a branch: https://github.com/bbito/node-mysql/blob/test-datestrings-options/test/integration/connection/test-datestrings-options.js
Just let me know if I should merge into master.

bbito:
Latest commit in test-datestrings-options branch includes 'mis-matched' tests - that is supplying JavaScript Date objects to the module when `dateStrings:` is truthy or 'date' and conversely supplying strings when `dateStrings:` is falsy. Although I don't think it is clearly documented, it seems that is is by design that this behavior is supported.

lib/protocol/packets/RowDataPacket.js
bbito: RE: Requested tests, @dougwilson could you take a quick look at the more exhaustive tests on my branch: https://github.com/bbito/node-mysql/blob/test-datestrings-options/test/integration/connection/test-datestrings-options.js and let me know if I should merge that into this PR?

dougwilson: Hi @bbito sorry I haven't had the time to revisit this PR just yet. As for additional tests, I'm all for it :) If you are making tests just as general additions to test the functionality that already exists, that is awesome! It would be nice to have it as a separate PR, ideally, but you can add it to this PR if you like.

bbito: As it stands the new test file will fail without the rest of the PR, so I'll merge it for now and if the PR is rejected and you want the test for the current code just let me know and I'll strip out the `dateStrings: 'date'` stuff and submit it as a separate PR.

dougwilson:
I swear I though I replied here at some point in the last few days, but I guess I'm just making stuff up in my mind :) Anyway, I need to take a look. Seems every time I turn around, more stuff is added to this PR, so please bear with me in trying to look it over :)

lib/protocol/packets/RowDataPacket.js
```diff
@@ -62,7 +62,11 @@ function typeCast(field, parser, timeZone, supportBigNumbers, bigNumberStrings,
     case Types.NEWDATE:
       var dateString = parser.parseLengthCodedString();
       if (dateStrings) {
-        return dateString;
+          if (field.type === Types.DATE) {
```
bbito: I'm happy to include it. When I first started poking into this module I found `NEWDATE` in the constants: https://github.com/mysqljs/mysql/blob/master/lib/protocol/constants/types.js , but I couldn't find it in mysql docs. Today I find 2 references to it in current docs: http://dev.mysql.com/doc/connector-net/en/connector-net-ref-mysqlclient-mysqlcommandmembers.html 

> Newdate: Obsolete. Use Datetime or Date type.

http://dev.mysql.com/doc/connectors/en/connector-net-ref-mysqlclient-mysqlcommand.html

> Newdate: Obsolete. Use Datetime or Date type.

So unless there is a decision for this module to remove support for this obsolete Member/Type and it is removed from the `switch` block in this file, I think I should amend this PR to support it.

bbito:
@dougwilson Is there anything more you want for this PR?

lib/protocol/packets/RowDataPacket.js
bbito: `Types.NEWDATE` support added with Oct. 7 2016 commit: https://github.com/mysqljs/mysql/pull/1481/commits/975b2ae9e21eb665c02a4c0e5e941bb16bf8b1da

cfilby:
Is there anything I could do to assist with this? Seems like the PR is just about done, but I'm definitely interested in this functionality and would be happy to help out.

dougwilson:
This is on my backlog, but is low priority since it doesn't add any functionality that is not already in this module; it simply adds some sugar to what you can already do using a custom typecast.

bbito:
@dougwilson in my scenario I don't think I have the ability to pass a custom typecast because I'm not using this module directly, but rather through Strongloop/Loopback: https://github.com/strongloop/loopback. 
In this scenario I can pass connection options through the \server\datasources.json configuration file, but I haven't found another way to customize a connection in Loopback.

dougwilson:
Hi @bbito, you may also want to make a PR to that wrapper in parallel to this PR, because the typecast is a very essential feature. I add things to this module from the perspective of using this module; if you are using a wrapper and it's forcing you to alter the module it's wrapping just to use it, you may want to use a different wrapper.

dougwilson:
Hi @bbito thanks for all your hard work. I am working on merging this now, for inclusion in the next release. The API would be an array similar to `debug` rather than just a string, thus `dateStrings: 'DATE'` would become `dateStrings: ['DATE']`, which allows people to fine tune whatever types they want casted or not for whatever reason.

PR #1496, opened 2016-08-12
Explain pooling a bit more

PanMan:
I think Pooling should be a bit more documented: This is a tiny first step at that

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -304,6 +304,8 @@ Unlike `end()` the `destroy()` method does not take a callback argument.
 
 ## Pooling connections
 
+Pooling is a connection manager, which handles the connections for your queries, and reconnects if needed. If you have a longer running process, you probably want to use this instead of using connection directly.
+
 Use pool directly.
 ```js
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -304,6 +304,8 @@ Unlike `end()` the `destroy()` method does not take a callback argument.
 
 ## Pooling connections
 
+Pooling is a connection manager, which handles the connections for your queries, and reconnects if needed. If you have a longer running process, you probably want to use this instead of using connection directly.
```
dougwilson: Connection pooling is a very generic concept. It looks like the first sentence is just giving it another name. Perhaps instead of that, we can provide a link to https://en.wikipedia.org/wiki/Connection_pool or somewhere similar that explains the concept of connection pooling?

dougwilson:
Since I didn't really hear back, I went ahead and pushed a change to add the link to that Wikipedia article. Please feel free to make a PR if you think there is anything that needs to be added to the README that the Wikipedia article does not touch on.

Issue #1505, opened 2016-08-26
add `.query` to poolNamespace


alsotang:
Now poolNamespace can only use with

```js
var pool = poolCluster.of('SLAVE*', 'RANDOM');
pool.getConnection(function (err, connection) {});
pool.getConnection(function (err, connection) {});
pool.query // !!!not a function
```

Its behavior is some different from `Pool`. With `Pool` we can use `.query` directly.

```js
var pool  = mysql.createPool({
  connectionLimit : 10,
  host            : 'example.org',
  user            : 'bob',
  password        : 'secret',
  database        : 'my_db'
});

pool.query('SELECT 1 + 1 AS solution', function(err, rows, fields) {
  if (err) throw err;

  console.log('The solution is: ', rows[0].solution);
});
```

dougwilson:
I don't see any reason why not to add it :) Feel free to work up a pull request, otherwise I can get around to it at some point :)!

alsotang:
OK, since you think it's ok, I would work for it this weekend. : )

merarischroeder:
I recommend forcing the use of Promise

```
static getConnection() {
	db.initialisePool();
	return new Promise((resolve, reject) => {
		db.pool.getConnection((err, connection) => {
			if (err) {
				connection.release();
				reject(err);
			}
			else {
				resolve(connection);
			}
		});
	});
}

static query(connection, query, params) {
	return new Promise((resolve, reject) => {
		connection.query(query, params, function (err, rows) {
			if (err) {
				if (!inTransaction) {
					connection.release();
				}
				reject(err);
			}
			else {
				resolve(rows);
			}
		});
	});
}

queryPromise(qry) {
	var state = {
		connection: null,
	};
	return db.getConnection().then((connection) => {
		state.connection = connection;
		return db.query(state.connection, query.queryString, query.queryParams, true);
	}).then((data) => {
		state.connection.release();
		return data;
	}).catch((error) => {
		state.connection.release();
		throw error;
	});
}
```

alsotang:
The usage of Promise seems another topic than poolNamespace @merarischroeder

merarischroeder:
I don't think you need a committee to return a Promise from a new function. You can have both a Promise and non-promise API for the .query function. 

Whether or not the final internal implementation uses Promises or not, is beside the point. But, why not use this as an opportunity to explore that?

Public Access:

- .query(sql, callback) - wraps .queryPromise
[...]

PR #1506, opened 2016-08-26
Add .query to PoolNamespace. 

alsotang:
Refactor PoolNamespace#_getClusterNode but still backward-compatible

original issue: https://github.com/mysqljs/mysql/issues/1505

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -467,6 +467,7 @@ poolCluster.of('*').getConnection(function (err, connection) {});
 var pool = poolCluster.of('SLAVE*', 'RANDOM');
 pool.getConnection(function (err, connection) {});
 pool.getConnection(function (err, connection) {});
+pool.query(function (err, result) {});
 
 // close all connections
 poolCluster.end(function (err) {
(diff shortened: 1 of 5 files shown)


alsotang:
now we can use cluster this way:

```js
 var pool = cluster.of('SLAVE*', 'ORDER');

  pool.query('SELECT 1', function (err, rows) {}) // works fine!!
```

alsotang:
travis ci fails, but seems a old test result in that.

dougwilson:
Nice, thanks, @alsotang! Don't worry about that failing test; I'll get it fixed today :)

dougwilson:
Sorry for the long delay, @alsotang ! I was able to fix the flaky tests, and so the PR is now passing the tests. I do see that the PR is marked failing as well because it caused the coverage percentage to decrease. Looking at the report, it looks like it's because in the new `PoolNamespace.prototype.query`, both of the `if (err)` code paths are not tested.

dougwilson:
lol @alsotang I just popped into this PR to see if I could just add the tests & merge to help out :) I see you just put more more commits. Please ping me when you're ready for me to take a look :)

alsotang:
I already add tests for the error handler. Seems now the `PoolNamespace.js` is 100% cover.

PR #1508, opened 2016-08-28
Added PoolClusterFarm - Need Help With Stored Procedures!

happilymarrieddad:
The PoolClusterFarm module allows a user to have many masters with many slaves. This allows for UNLIMITED scaling with any application. I created this module because I need it for another project I'm working on and I thought it would be better to add it for others instead of just creating a wrapper or something. It probably could use some additional options for some edge cases. Thanks!
## PoolClusterFarm

PoolClusterFarm allows for a complete replicated MySQL farm to be handled with the mysql module. For more information on MySQL replicating, please navigate to http://dev.mysql.com/doc/refman/5.7/en/replication.html. Basically, this module allows an unlimited amount of scaling with 1 to Many Masters and 0 to Many Slaves. In order to use this module, you must set up replication masters and slaves using the tutorial above.

```
var mysql = require('mysql');
var poolClusterFarm = mysql.createPoolClusterFarm();

poolClusterFarm.add('MASTER0','master',{database:'test',user:'admin',password:'somepassword',host:'10.0.0.1'});
poolClusterFarm.add('MASTER1','master',{database:'test',user:'admin',password:'somepassword',host:'10.0.0.2'});

poolClusterFarm.add('SLAVE0','slave',{database:'test',user:'admin',password:'somepassword',host:'10.0.0.3'});
poolClusterFarm.add('SLAVE1','slave',{database:'test',user:'admin',password:'somepassword',host:'10.0.0.4'});
poolClusterFarm.add('SLAVE2','slave',{database:'test',user:'admin',password:'somepassword',host:'10.0.0.5'});
poolClusterFarm.add('SLAVE3','slave',{database:'test',user:'admin',password:'somepassword',host:'10.0.0.6'});

poolClusterFarm.query('SELECT id,first FROM users WHERE id = 1',function(err,rows) { console.log(err);console.log(rows) });
poolClusterFarm.query('UPDATE users SET first = "testing" WHERE id = 1',function(err,rows) { console.log(err);console.log(rows) });
```
### PoolCluster options

```
var clusterConfig = {
  debug:true
};

var poolClusterFarm = mysql.createPoolClusterFarm(clusterConfig);
```


happilymarrieddad:
@dougwilson Hey man, is this something that we can look at? I really need this for one of our production servers and I've tested it in a staging environment for quite a while. I believe it's pretty stable. Thanks!

Readme.md
```diff
@@ -525,6 +527,35 @@ resets any connection state (variables, transactions, etc.).
 Errors encountered during this operation are treated as fatal connection errors
 by this module.
 
+## PoolClusterFarm
+
+PoolClusterFarm allows for a complete replicated MySQL farm to be handled with the mysql module. For more information on MySQL replicating, please navigate to http://dev.mysql.com/doc/refman/5.7/en/replication.html. Basically, this module allows an unlimited amount of scaling with 1 to Many Masters and 0 to Many Slaves. In order to use this module, you must set up replication masters and slaves using the tutorial above.
```
dougwilson: Please wrap the lines to match the rest of the markdown :)

dougwilson: Also, make the URL a link.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+  this._numMasters = 0;
+  this._numSlaves = 0;
+  this._last_index = 0;
+  this.debug = config.debug || false
+}
+
+Util.inherits(PoolClusterFarm,EventEmitter);
+
+PoolClusterFarm.prototype.query = function query(sql,values,cb) {
+	var self = this
```
dougwilson: Would it be possible to make the indentation match that of all the other files?

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+
+	if (!this._numMasters || this._numMasters < 1) {
+		if (cb && typeof cb == 'function') { return cb('No servers assigned to PoolClusterFarm. Please verify...'); }
```
dougwilson: You may want to put a preamble to set the callback to a variable instead of repeating this detection over and over.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+		else { return values('No servers assigned to PoolClusterFarm. Please verify...'); }
```
dougwilson: Never pass strings as errors; please construct an `Error` object and pass that back as the first argument to the callback.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	if (typeof sql != 'string' || sql.length < 7) {
```
dougwilson: The `sql.length < 7` seems completely arbitrary. What is the purpose of that? Are you now allowing someone to send the valid SQL `BEGIN`?

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+		if (cb && typeof cb == 'function') { return cb('Invalid sql statement. Please verify...'); }
+		else { return values('Invalid sql statement. Please verify...'); }
+	}
+
+	if (sql.substr(0,6).toLowerCase() == 'select') {
```
dougwilson: imo this is too brittle to accept as it is. There is no guarantee that the user won't have whitespace at the beginning of the string (from template strings, for example), and I'm not sure this correctly handles multiple statements in the same query (and I don't see anything in here that prevents users from turning that on?).

Also, it seems like if someone runs the SQL `SELECT ... FOR UPDATE` they probably want that to happen on the master, not a slave? What else are we missing here?

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+		else { this._last_index++; }
+		if (this.debug) { console.log('Using server',(index_to_use+1),'of',(this._servers.length)); }
+		this._servers[index_to_use].pool.query(sql,values,cb);
+	} else {
+		var num = this._numMasters;
+
+		var updateMasters = function(i) {
+			self._servers[i].pool.query(sql,values,function(err,rows) {
+				if (!(--num)) {
+					if (cb && typeof cb == 'function') { return cb(err,rows); }
```
dougwilson: It looks like this looses all errors and only forwards the success/failure of the last master. I would expect that in most error situations, the errors happen faster than the successes (since there isn't much work done). This would seem to make it look like the query typically succeeds instead of fails due to this race condition.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+		id 			: 	server_name,
+		type 		: 	server_type,
+		pool 		: 	new Pool({config:poolConfig}),
+		online 	: 	1
+	})
+	this._calculateNumServers()
+}
+
+PoolClusterFarm.prototype._verifyConfig = function _verifyConfig(config) {
+	if (!config.user) { return 'User must be passed in to make a database connection'; }
```
dougwilson: Why this validation? I regularly interact with database that do not require a username and password. There is no requirement from MySQL to require these configurations.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+
+	}
+}
+
+PoolClusterFarm.prototype.add = function add(id,type,config,respond) {
+	if (!respond || typeof respond != 'function') { respond = function(err){ throw new Error(err) }; }
+	var server_name = typeof id == 'string' ? id : this._makeId();
+	var server_type = typeof type == 'string' && (type == 'master' || type == 'slave') ? type : 'master';
```
dougwilson: Just defaulting to `'master'` seems prone to typos. Any reason not to just have two methods: `.addMaster` and `.addSlave`?

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+	var server_config = {};
+	if (config && typeof config == 'object') {
+		server_config = config;
+	} else if (type && typeof type == 'object') {
+		server_config = type;
+	} else if (id && typeof id == 'object') {
+		server_config = id;
+	} else {
+		return respond('No server configuration passed in. Please look at the documentation.');
```
dougwilson: This is a developer error and from Node.js conventions, should be a sync throw, not provided to the callback. Also, should be a `TypeError` object, probably.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+	}
+	var err = this._verifyConfig(server_config);
+	if (err) { return respond(err); }
```
dougwilson: This is a developer error and from Node.js conventions, should be a sync throw, not provided to the callback. Also, should be a `TypeError` object, probably.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+	var poolConfig = new PoolConfig(server_config)
+	this._servers.push({
+		id 			: 	server_name,
```
dougwilson: Seems like a bunch of messed up whitespace here.

lib/PoolClusterFarm.js
```diff
@@ -0,0 +1,120 @@
...
+
+PoolClusterFarm.prototype._makeId = function _makeId() {
+	return Math.random().toString(36).substring(5);
+}
+
+PoolClusterFarm.prototype._setLocked = function _setLocked(locked) {
+	this._locked = locked ? true : false;
+}
+
+PoolClusterFarm.prototype.isLocked = function isLocked() {
```
dougwilson: Where is this used? I'm not seeing it used (or documented?) any where.

test/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
dougwilson: This file seems to be added in error, if you can remove it, please :)

happilymarrieddad:
Cool, thanks so much for your time, @dougwilson. I will take the time to do this right. Thanks!!

Readme.md
happilymarrieddad: ok got this done.

lib/PoolClusterFarm.js
happilymarrieddad: fixed

lib/PoolClusterFarm.js
happilymarrieddad: Ok, I fixed this. I didn't remove the set of them because I was thinking it was ok if somebody didn't pass in a callback. I'm not sure if it's the best but there are times I'll call the database and I don't really care if it's successful.

lib/PoolClusterFarm.js
happilymarrieddad: I fixed this. Thanks

lib/PoolClusterFarm.js
happilymarrieddad: Fixed it

lib/PoolClusterFarm.js
happilymarrieddad: You are right.. this is the primary reason I have been delaying fixing this PR. I think I have a solution you may like.

happilymarrieddad: Ok, it is super brittle so I do a match and if not use masters. The problem with this is it's possible for a query that would work on a slave to be sent to a master but will only happen if they are using multiple statements which is a rare occurrence. It shouldn't add too much extra processing to the masters compared to the slaves.

lib/PoolClusterFarm.js
happilymarrieddad: ok I monitor errors and immediately returns with the error and stops all other execution. I'm not certain this is the best solution but it does accomplish the goal.

lib/PoolClusterFarm.js
happilymarrieddad: You are correct. I've been meaning to do this but I haven't yet. I'll take care of it.

lib/PoolClusterFarm.js
dougwilson: But wouldn't that cause all the update/insert procedure calls also go to the slaves?

dougwilson: Haha, I guess I wouldn't be allowed to use this module then ;) I'm not sure if there are really any users who sit down and flush out a giant code base of hand written SQL mixed in their JavaScript, though. I imagine that there are really only two main type of production callers (1) those who use an ORM system that generates SQL, but the same ORM already understands reads/writes and is better off using the pool cluster for better control and (2) those who write stored procedures and call those from applications (which works really well when you have different programming languages interacting with the same database), which also is better off using pool cluster.

lib/PoolClusterFarm.js
happilymarrieddad: You are correct sir! Lol, stepped in that one. I will fix it.

lib/PoolClusterFarm.js
happilymarrieddad: fixed

lib/PoolClusterFarm.js
happilymarrieddad: oh I thought it made it easier to see what I'm doing. I'll change it.

lib/PoolClusterFarm.js
happilymarrieddad: I just removed the requirements.

lib/PoolClusterFarm.js
happilymarrieddad: Oh, i was going to lock on updates but decided that it should be handled by the db. I remove it.

happilymarrieddad:
Ok, I just need to solve the stored procedure issue and I think we're good to go. I'll throw up the new build on my staging servers for a few days just to see how it runs before I try anything else. Thanks!

lib/PoolClusterFarm.js
happilymarrieddad: @dougwilson hey man the only way I can think to do this is if someone is using mostly stored procedures then they're just going to have to run all masters. It will still be pretty fast. I can't think of a way for me to know what kind of query/queries are in the stored procedure. Do you know of any way to describe a stored procedure and then do logic based on the result without significantly slowing down the module? thanks!

dougwilson: I have no idea, which is why I'm putting it out here to get multiple people thinking on it :) It's not that different from using something like `LAST_INSERT_ID()`, where someone will need to perform one query that does `INSERT ...` and then needs to do a second query using `LAST_INSERT_ID()`. As far as I can tell, that's not currently possible with this PR either?

happilymarrieddad:
@dougwilson Hey man, I was thinking about adding Redis for caching queries to speed it up significantly. What do you think? Have you heard anything on some possible solutions for stored procedures?

test/package.json
happilymarrieddad: done

lib/PoolClusterFarm.js
dougwilson: How would it be a greater benefit? Stored procedures benefit from being split across slaves/masters just as much as ad-hoc SQL does.

happilymarrieddad: Well, you would have one object, pool.query(), and it would allow for you to have unlimited scaling. For example, if you had a large app that had millions of users, one database server probably can not handle that much traffic. That being said, a cluster of master servers can handle all those requests pretty easily and it would scale. Sorry if I'm not being clear.

happilymarrieddad: For example, any SQL sent would be sent to all the servers.. Basically, this is a way to do replication without having to have only 1 master server if that makes sense.

dougwilson:
Yea, I'm starting to feel like this PR should not be merged, because it is very opinionated on how to set things up, seems hostile towards stored procedures / multiple statements / other things, and even worse, if we add it here, then no one could use a different backend like the `mysql2` module which provides binary protocol support. I don't see this module using anything internal to the `mysql` module, and it seems like it would greatly benefit from being published to npm as a standalone module, such that people can back the module with `mysql` or `mysql2` or something else easily. @sidorares ?

happilymarrieddad:
I think you're probably right. Thanks for all your time @dougwilson Doug Wilson!!!

dougwilson:
Cool, yea. Let me know if you need any help trying to make it a standalone module that wraps this / `mysql2` :)

PR #1534, opened 2016-10-04
Typo fix in error message

vvm-broadsoft:
Extra "n" in "Server does not support secure connnection" error.


dougwilson:
Awesome, thanks for the catch!

PR #1543, opened 2016-10-13
Create Demo-Contribution

Rachana12Sag:


dougwilson:
Hi @Rachana12Sag thank you very much for your contribution, but I think this may be from https://en.wikipedia.org/wiki/MySQL and it looks like the text content on that page is not licensed under something that is compatible with our license (and it doesn't look like the work had been properly attributed to the original source per the license to use the text).

Issue #1272, opened 2015-11-05
Ability to log all queries


q42jaap:
I want to plugin in nodejs' `debug` into node-mysql, in order to log all the queries. I'd rather not wrap all the calls to node-mysql, because of the formatting and the many combinations of arguments.

I'm using bluebird to make promises so I can't really use the return value of `connection.query()`. `connection.queryAsync()` doesn't give me access to the return object.

It would be nice if there was a way to plug it in, maybe an event on the connection/pool object. That way I don't have to think about writing the code that does the logging.

dougwilson:
Hi @q42jaap , that seems like a pretty good idea. Any proposal on what would be a good way to go about it that would satisfy your use-case?

q42jaap:
Hi, sorry it took a while for me to respond.
I would think a `Connection` level event system seems a good idea.

```
connection.on('query', function(query) { console.log(query.sql) });
```
as an added bonus, it may be nice to also add it to a `Pool`:

```
pool.on('query', function(query) { console.log(query.sql) });
```

Something along these lines?

ayanth:
+1

dougwilson:
Perhaps the following code may work?
```js
connection.on('enqueue', function (sequence) {
  if (sequence instanceof mysql.Sequences.Query) {
    console.log(sequence.sql);
  }
});
```

The response didn't actually lay-out the use-case, only really what you envisioned the API to be. Can you please layout the use-case for the event? For example, why emit queries both in the `enqueue` event as well as again on a `query` event?

q42jaap:
Hi @dougwilson,
I did not know this! This may very well work for my use case. I just want to see all the query's while developing. Enqueue is probably exactly what I'm looking for.

As a last thing, does the pool also have this enqueue event?

dougwilson:
> As a last thing, does the pool also have this enqueue event?

It does, but it has a different meaning: it means a connection request has been enqueued. If you want to monitor from the pool-level, just combine the `connection` event on the pool plus the `enqueue` event on the connection:
```js
pool.on('connection', function (connection) {
  connection.on('enqueue', function (sequence) {
    if (sequence instanceof mysql.Sequences.Query) {
      console.log(sequence.sql);
    }
  });
});
```

q42jaap:
This is super helpful, thanks a lot!

dougwilson:
No problem!

Telokis:
[...]

PR #1559, opened 2016-10-29
Readme.md: adds section about query logging

rovolution:
First off, I wanted to say "thank you" for creating and maintaining this great module!

Recently, I ran into a use case where I wanted to log all the raw queries being made by a connection pool.

I had to dig around the Issues tab a bit until I landed upon [this](https://github.com/mysqljs/mysql/issues/1272#issuecomment-170144701). It was close to the solution I needed (I dug into the source code and saw that the `Sequence` object was not being bound to the main export Object, so ended up removing the `mysql.Sequences.Query` line and it worked like a charm).

I think having this little snippet in the README itself would help future users quickly find out how to do this. I figured it made most sense as a sub-section within the `Debugging and reporting problems` section.

Please let me know if should adjust anything or if adding this to the docs does not make sense. 

Thanks!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1327,6 +1327,20 @@ will have:
 * As much debugging output and information about your environment (mysql
   version, node version, os, etc.) as you can gather.
 
+### Logging All Queries from a Connection Pool
+
+To log all the queries from a connection pool, bind to the `enqueue` event on each connection and check for the `mysql` property on the `sequence` object:
+
+```js
+pool.on('connection', function (connection) {
+  connection.on('enqueue', function (sequence) {
+    if (sequence && sequence.sql) {
+      console.log(sequence.sql);
+    }
+  });
+});
+```
+
 ## Contributing
 
 This project welcomes contributions from the community. Contributions are


dougwilson:
Hi @rovolution thanks for the PR! That section of the README is specifically about debugging & reporting problems with this module to us, not just in general. When you use `debug: true` it will already output the queries, so adding a subsection to do the same thing is pretty redundant. Perhaps there may be a better location in the README?

rovolution:
@dougwilson perhaps moving it to the [Pooling Connections](https://github.com/mysqljs/mysql#pooling-connections) section?

Another thing we could potentially do is add a new section that discusses different ways to access the raw MySQL string. For example, how to access it for a single query, a single connection, a pool of connections, etc...

Let me know what you think. Thanks!

Arnoldnuo:
@rovolution Thank you very very very very very very much !

rovolution:
@dougwilson any updates on this? Thanks!

dougwilson:
I haven't giving it much thought beyond that it definitely doesn't belong in the current section, so there isn't any changes to really merge at this time. I know you gave a suggestion, but that doesn't seem right, either. I also noticed your example uses undocumented feature, so I think it would be important that everything in your example is actually documented before added it, or that just raises more questions.

rovolution:
sounds good. ill go ahead and close this then. Thanks!

PR #1571, opened 2016-11-14
Improved the logic of Pool's changeUser.

ifsnow:
I don't think `connection.changeUser()` would be used frequently in the real world.
But, An unnecessary routine for this is tested whenever `pool.getConnection()` is called.

This PR is aimed at reduction in this complexity. As a result, `pool.getConnection` works 3~4 times faster.

In addition, a small bug was fixed. (lib/ConnectionConfig.js)

### DIFF
--- a/lib/protocol/sequences/ChangeUser.js
+++ b/lib/protocol/sequences/ChangeUser.js
@@ -26,10 +26,12 @@ ChangeUser.prototype.start = function(handshakeInitializationPacket) {
     charsetNumber : this._charsetNumber
   });
 
-  this._currentConfig.user          = this._user;
-  this._currentConfig.password      = this._password;
-  this._currentConfig.database      = this._database;
-  this._currentConfig.charsetNumber = this._charsetNumber;
+  var currentConfig = this._currentConfig;
+  currentConfig.user          = this._user;
+  currentConfig.password      = this._password;
+  currentConfig.database      = this._database;
+  currentConfig.charsetNumber = this._charsetNumber;
+  currentConfig.changedUser   = true;
 
   this.emit('packet', packet);
 };
--- a/lib/ConnectionConfig.js
+++ b/lib/ConnectionConfig.js
@@ -57,7 +57,7 @@ function ConnectionConfig(options) {
 
   // Set the client flags
   var defaultFlags = ConnectionConfig.getDefaultFlags(options);
-  this.clientFlags = ConnectionConfig.mergeFlags(defaultFlags, options.flags);
+  this.clientFlags = ConnectionConfig.mergeFlags(defaultFlags, this.flags);
 }
 
 ConnectionConfig.mergeFlags = function mergeFlags(defaultFlags, userFlags) {
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -19,8 +19,7 @@ function Pool(options) {
   this._closed               = false;
 }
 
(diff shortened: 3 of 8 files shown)


lib/protocol/sequences/ChangeUser.js
```diff
@@ -7,29 +7,28 @@ module.exports = ChangeUser;
...
-  this._currentConfig.user          = this._user;
-  this._currentConfig.password      = this._password;
-  this._currentConfig.database      = this._database;
-  this._currentConfig.charsetNumber = this._charsetNumber;
+  var currentConfig = options.currentConfig;
+  currentConfig.user          = options.user;
+  currentConfig.password      = options.password;
+  currentConfig.database      = options.database;
+  currentConfig.charsetNumber = options.charsetNumber;
+  currentConfig.modified      = true;
```
dougwilson: I ran a v8 regression against this really quick and this is altering the hidden class during runtime. Please ensure that the objects build the correct hidden class from the get-go and don't alter it through runtime

dougwilson:
> As a result, pool.getConnection works 3~4 times faster

Please provide the code you used to benchmark this and provide the before / after benchmarks. We can get your benchmark added to our benchmarks folder :) !

dougwilson:
> I don't think connection.changeUser() would be used frequently in the real world.

What is this based on? It was a request by the community and I use it all the time. This claim seems dubious. Can you provide some data to back up this assumption?

dougwilson:
I see you're setting the `modified` property to `true` to trigger a change user, but I don't see anywhere you are changing it back? Wouldn't that cause a change user on every single pool connection from that point forward? That seems like a bug.

dougwilson:
> Wouldn't that cause a change user on every single pool connection from that point forward? That seems like a bug.

Nevermind, it just uses that to short-circuit the real checks. Still seems a bug if the purpose of this is performance, which means that just doing one change user will kill the performance for the future check outs since the `modified` is never reset back to `false`.

lib/protocol/sequences/ChangeUser.js
```diff
@@ -7,29 +7,28 @@ module.exports = ChangeUser;
 Util.inherits(ChangeUser, Sequence);
 function ChangeUser(options, callback) {
   Sequence.call(this, options, callback);
-
-  this._user          = options.user;
```
dougwilson: The packets are publicly exposed. Even though these are `_`-prefixed properties, historically when we change these people break; please keep all the old properties, based on our past experiences.

ifsnow:
@dougwilson
Thanks for your quick feedback.

## benchmark

[code](https://gist.github.com/ifsnow/fff8af1cb3e7db1ba57c56369622c27f)

I'm sure it's a micro-optimization. This may not be important.
However, it's easy to do. I think it's just a difference of view. :)

## changeUser usage
In my experience, Changing user of `Pool` is too inefficient.

```
Pool with user A -> pool.getConnection() -> changeUser to user B (user) 
-> connection.release() -> pool.getConnection() -> changeUser to user A (system)
```

I prefer to create separate pools with user A and user B if it needs.
Especially, if it's on production mode.

I think it's more likely that this function will not be used in common cases.
Do you need to change user of `Pool` frequently?

## reset modified field
It's a bug. fixed. and `modified` changed to `changedUser`.

## keep the old properties
I couldn't think of it. It's been rollbacked.

PR #1578, opened 2016-11-20
docs: add note about callback parameters

KernelDeimos:


dougwilson:
I think adding this warning in the location that it is actually shows how dangerous it is that we provide code examples, as it seems as soon as there is a code example, people will copy and paste it and never read further. I think it would be better to remove the code examples all together from this section for the query and direct them straight to the query section to learn how to make them instead of this warning block.

dougwilson:
In fact, it doesn't even look like this is the first place in the README where a user would encounter the `.query` method when reading from top to bottom. It seems like if this warning block were to go anywhere, it would most below at the first occurrence of the `.query` in the docs, right?

KernelDeimos:
The first code example has all three fields, so it wouldn't cause such a mistake. Although I definitely agree with what you said before that, about the danger of those code examples - I propose replacing this
`connection.query('SELECT 1', function(err, rows) { ...`
with simply
`connection.query( ... );`
and then do something similar for other examples.

dougwilson:
> The first code example has all three fields, so it wouldn't cause such a mistake.

Gotcha. So then, what makes this code example worth of the warning? Would you not need to place that warning at every single other `.query` code that also does not have three arguments? It just seems arbitrary.

Why not just go through every single `.query` and add all three arguments there (be sure that the `.query` is the `conn.query` one--I believe there are a couple others that are not quite the same?) include comments inline in the code to note what each of those arguments mean in every single code snippet? I think that would be a lot more useful than just putting the warning on this one `.query` when many, many others look like they would cause the exact same confusion.

If you do `connection.query( /* query parameters here */ );`, then make sure you put a new section to explain what `/* query parameters here */` means and probably link to it from everywhere or something, as a comment is, I think, a lot harder to remember than argument names, especially since a lot of people coming to this module are new to Node.js and will likely forget that the first argument is `error`, basically the inverse of the `fields` issues you had.

KernelDeimos:
I agree, that would definitely be more useful.

This is how I did the first one - any criticism on this format?
```js
connection.query('SELECT 1', function(
  err,   // Error object, or null if successful
  rows,  // List of rows
  fields // Ordered list of field names
) {
  // connected! (unless `err` is set)
});
```

dougwilson:
Unless there is a reason not to, I would say just use the exact format that is used in the "Performing queries" section.

dougwilson:
Hi @KernelDeimos so since I never heard anything back or any new commit here, I made the changes that I think reflected our conversion. If you think there are additional changes, please feel free to open another PR!

PR #1591, opened 2016-12-01
New Pool options and refactoring.

ifsnow:
I hope `Pool`'s features will get better like the others, so I added several options such as `queueWaitTimeout`, `pingCheckInterval`, `startConnections`, `minSpareConnections`, `maxSpareConnections`, `spareCheckInterval`
([See Readme.md](https://github.com/ifsnow/mysql/blob/improve-pool-options/Readme.md#pool-options))

## Efficient code
`Pool` will work 10~50 times faster with new options.
  
|  | New ([benchmark code](https://gist.github.com/ifsnow/c4672b09a12e2a34c05d3c5ffd9d2f1e)) | Original ([benchmark code](https://gist.github.com/ifsnow/609c393e007daaad5f49bd2c60d32509))|
| --- | --- | --- |
| Executions per second | 1,292,685 | 22,415 |
  
## Easy-to-read code
I've tried to make the code easier to read and maintain for the following contributors, and tidy up wrong and unused logics.

## Backwards compatibility code
All defaults for new options are off. All tests are passed, so existing users can use it right now.
In my opinion, It's recommended that they use the new options as other pools do.

Here is my config below.

```  
{
  connectionLimit : 50,
  queueWaitTimeout : 10000, // Same as acquireTimeout.
  pingCheckInterval : 10000, // The connection used in 10 seconds is reused without ping check.
  startConnections : 10, // 10 connections are created when the pool is started.
  minSpareConnections : 10, // 10 spare connections should be kept in the pool at all times.
  maxSpareConnections : 20, // No more than 20 spare connections.
  spareCheckInterval : 300000 // Check the spare connections every 5 minutes.
}
```

If you agree on my opinion, you can change the default value of new options. or, I think you could add a comment at Readme.md like this.
  
```
`pingCheckInterval`: ........ (Default: `0`, Recommendation: `10000`)
```

***

I hope this helps.
Please let me know if you have any concerns or advice.

### DIFF
--- a/test/unit/pool/test-destroy-connection.js
+++ b/test/unit/pool/test-destroy-connection.js
@@ -11,10 +11,10 @@ server.listen(common.fakeServerPort, function (err) {
   pool.getConnection(function (err, connection) {
     assert.ifError(err);
 
-    assert.strictEqual(connection, pool._allConnections[0]);
+    assert.strictEqual(connection, pool._connectionManager._allConnections._map[1]);
     connection.destroy();
 
-    assert.strictEqual(pool._allConnections.length, 0);
+    assert.strictEqual(pool._connectionManager._allConnections.size(), 0);
     assert.ok(!connection._pool);
 
     assert.doesNotThrow(function () { connection.release(); });
--- a/Readme.md
+++ b/Readme.md
@@ -389,12 +389,27 @@ constructor. In addition to those options pools accept a few extras:
 * `queueLimit`: The maximum number of connection requests the pool will queue
   before returning an error from `getConnection`. If set to `0`, there is no
   limit to the number of queued connection requests. (Default: `0`)
+* `queueWaitTimeout`: The maximum number of milliseconds that the pool will wait
+  for a connection when there are no available spare connections. If set to `0`,
+  there is no limit. (Default: `0`)
+* `pingCheckInterval`: The number of milliseconds to indicate how often to check
+  the validity of the connection. If set to `0`, checks whether or not the connection
+  to the server is working every time. Setting this value in high-traffic can
+  improve performance. (Default: `0`)
+* `startConnections`: The initial number of connections that are created when the
+  pool is started. (Default: `0`)
+* `minSpareConnections`: The minimum number of spare connections that should be
+  kept in the pool at all times. If set to `0`, it doen't work. (Default: `0`)
+* `maxSpareConnections`: The maximum number of spare connections that should be
+  kept in the pool at all times. If set to `0`, it doen't work. (Default: `0`)
+* `spareCheckInterval`: The number of milliseconds to indicate how often to check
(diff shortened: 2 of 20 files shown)


dougwilson:
Thanks a lot @ifsnow ! Is it possible to break this up into multiple PRs instead of one massive one? At the very lest to break it up into many different commits in the PR? It's a _lot_ to really look through and understand the edge cases that will break people if this is applied without a major version bump.

test/unit/pool/test-destroy-connection.js
```diff
@@ -11,10 +11,10 @@ server.listen(common.fakeServerPort, function (err) {
   pool.getConnection(function (err, connection) {
     assert.ifError(err);
 
-    assert.strictEqual(connection, pool._allConnections[0]);
+    assert.strictEqual(connection, pool._connectionManager._allConnections._map[1]);
```
dougwilson: Even though this is an underscore-prefixed property, people really do use it, as I've seen throughout Stackoverflow and issues here. We can't change this without a major at this point. Hopefully it's possible to implement the idel timeout stuff without making this change.

ifsnow: It's a sad situation. I'll work to provide important old properties.

dougwilson: At this point in the module, probably every single property is important in some way. I would err on the site of just not changing them if you don't really need to.

ifsnow: I really understand. but, I think it's uncommon to use those properties such as `Pool._acquiringConnections` directly. It seems that it's not easy and efficient to work with all the existing ones.

I thought shortly.
```js
pool.createPool({
    ...,
    keepOldProperties : true
});
```
If the user set this, 4 properties(_acquiringConnections, _allConnections, _freeConnections, _connectionQueue) are updated. Isn't it enough?

ifsnow: This PR does not make sense if keeping all properties is essential. I'm a little worried about what to do. Unfortunately, I think I'd better close this.

dougwilson: @ifsnow I think a lot of the pool optimizations are worth it, and certainly can make it into a major version of this module. The only real concern I have is that because this is just one giant PR, making good (but breaking) changes _and_ making good but non-breaking changes, together, I have no option but to delay all the features in this PR until the next major. If you can simply break it up into the different pieces as multiple PRs, we can easily accept the easy ones now and discuss the breaking ones better and decide on the release of a new major version.

ifsnow: @dougwilson I really know how difficult it's to review many changes. I'm going to think about how I can break it up but, it's not easy. That's because `PoolConnectionManager` has all new functions. I wonder if it's possible to split this into multiple PRs easily.

PR #1614, opened 2016-12-30
Update Protocol.js

SystemDisc:
```
/path/to/project/node_modules/mysql/lib/protocol/Protocol.js:152
    .on('packet', function(packet) {
    ^

TypeError: Cannot read property 'on' of undefined
    at Protocol._enqueue (/path/to/project/node_modules/mysql/lib/protocol/Protocol.js:152:5)
    at Protocol.handshake (/path/to/project/node_modules/mysql/lib/protocol/Protocol.js:52:41)
    at Connection.connect (/path/to/project/node_modules/mysql/lib/Connection.js:136:18)
    at Object.<anonymous> (/path/to/project/dist/server/index.js:62122:12)
    at __webpack_require__ (/path/to/project/dist/server/index.js:27:30)
    at Object.<anonymous> (/path/to/project/dist/server/index.js:62089:12)
    at __webpack_require__ (/path/to/project/dist/server/index.js:27:30)
    at Object.<anonymous> (/path/to/project/dist/server/index.js:47017:19)
    at __webpack_require__ (/path/to/project/dist/server/index.js:27:30)
    at Object.<anonymous> (/path/to/project/dist/server/index.js:63784:13)
````

### DIFF
--- a/lib/protocol/Protocol.js
+++ b/lib/protocol/Protocol.js
@@ -148,14 +148,17 @@ Protocol.prototype._enqueue = function(sequence) {
   sequence
     .on('error', function(err) {
       self._delegateError(err, sequence);
-    })
+    });
+  sequence
     .on('packet', function(packet) {
       Timers.active(sequence);
       self._emitPacket(packet);
-    })
+    });
+  sequence
     .on('end', function() {
       self._dequeue(sequence);
-    })
+    });
+  sequence
     .on('timeout', function() {
       var err = new Error(sequence.constructor.name + ' inactivity timeout');
 
@@ -164,7 +167,8 @@ Protocol.prototype._enqueue = function(sequence) {
       err.timeout = sequence._timeout;
 
       self._delegateError(err, sequence);
-    })
+    });
+  sequence
     .on('start-tls', function() {
       Timers.active(sequence);
       self._connection._startTLS(function(err) {


SystemDisc:
Nevermind, this appears to be a Node bug. https://github.com/nodejs/node/issues/10535

SystemDisc:
https://github.com/nodejs/node/issues/10535

AdriVanHoudt:
wait isn't the issue not that the sequence is not created with `new`?

SystemDisc:
Although, looking through the code again, it seems `new` is used on `Sequences.Handshake` on line 52 of Protocol.js, and it seems `Sequence.call(this, options, callback);` in Handshake.js and `EventEmitter.call(this);` in Sequence.js should be enough.... I'm not sure why this error is happening...

SystemDisc:
@lpinca, since you were so helpful in https://github.com/nodejs/node/issues/10535, do you mind taking a look at this?

dougwilson:
Hi @SystemDisc I'm not sure what you're seeing, either. Perhaps if you could provide us the details in order to reproduce the error, we could debug the issue?

lpinca:
I can only think of this zone.js issue https://github.com/angular/zone.js/issues/508.
We had a similar bug report in ws.

SystemDisc:
If I change the version for zone in package.json to 0.7.4, everything works without problems. It seems to me that this is related to https://github.com/angular/zone.js/issues/508

Thanks so much for your help.

dougwilson:
Ah, good catch. Sounds like updating that dep fixed the issue :)

PR #1630, opened 2017-01-26
fields added to package.json

lkisac:
I've added/modified a few fields in the package.json file: _repository_, _keywords_ and _bugs_, so they validate with http://package-json-validator.com.  

Let me know if these changes are ok and if you would like anything modified with them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,13 @@
     "type": "git",
     "url": "http://github.com/mysqljs/mysql"
   },
+  "keywords": [
+    "mysql",
+    "sql",
+    "database",
+    "query"
+  ],
+  "bugs": "http://github.com/mysqljs/mysql/issues",
   "dependencies": {
     "bignumber.js": "3.1.2",
     "readable-stream": "1.1.14",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,10 @@
     "Diogo Resende <dresende@thinkdigital.pt>",
     "Nathan Woltman <nwoltman@outlook.com>"
   ],
-  "repository": "mysqljs/mysql",
+  "repository": {
+    "type": "git",
+    "url": "http://github.com/mysqljs/mysql"
+  },
   "dependencies": {
     "bignumber.js": "3.1.2",
     "readable-stream": "1.1.14",


dougwilson:
Thanks for the effort, but unfortunately unnecessary. Npm automatically populates the github info from the short string field, which is nicer and the new npm search no longer uses the keywords array in the search.

You may want to report the issues to the validator tool.

lkisac:
Ok, I will check with them.  Thanks.

dougwilson:
Awesome! Npm itself uses https://npmjs.org/package/read-package-json to read the file, so that would probably be a good reference implementation. Npm itself will throw errors for invalid fields, warnings for missing / wrong fields on publish or package read and this module is currently clean for that as well, so I don't understand what the validator is validating against, as I would assume it should work like npm.

dougwilson:
Looking through the GitHub repo for that validator tool, there at lots of open issues regarding the issues I noted above regarding incorrect validation for npm packages like https://github.com/gorillamania/package.json-validator/issues/29 and https://github.com/gorillamania/package.json-validator/issues/44

lkisac:
I should note that the only non-valid field was the repository url.  keywords and bugs were only at warning level.  There is also a recommendation for the 'homepage' field, which I left alone.

package-json-validator
**Results:**
```
{
  "valid": false,
  "errors": [
    "Url not valid for repository: mysqljs/mysql"
  ],
  "warnings": [
    "Missing recommended field: keywords",
    "Missing recommended field: bugs"
  ],
  "recommendations": [
    "Missing optional field: homepage"
  ]
}
```

Looking at the open issues you provided, it seems that the normalization rules are different between https://www.npmjs.com/package/normalize-package-data and https://www.npmjs.com/package/read-package-json

dougwilson:
The repository field in our repo is definitely valid. I liked you to the issue reporting that the validator is incorrectly thinking it is invalid. Also the bugs and homepage URLs are auto populated from the repository short string to their GitHub addresses, which is why we don't need to specify them yet they show up on the npmjs.com site and in the npm repository just fine, so there is nothing to change here, I think you just accidentally stumbled across some bugs in that validator tools, which is why I suggested to file issues there and hopefully they'll get it fixed.

LinusU:
Yes it's valid, it's the validator that is a bit outdated. I hope to get a weekend to give it some love soon :)

Issue #483, opened 2013-05-13
Query Execution Time


jbrooksuk:
Is there anyway to get how long the query took to execute?

sidorares:
@dougwilson I actually think command line client measures time itself. There is no extra requests sent over network to get this information and there is no time in any of result packages

HaykoKoryun:
+1 for this issue
http://stackoverflow.com/questions/26141129/nodejs-mysql-measure-query-execution-time

PR #1645, opened 2017-02-14
Added the query execution time in query()'s callback.

ifsnow:
This is for #483 

Added to pass `queryTime` as the fourth argument of `query()`'s callback.

```js
connection.query('SELECT SLEEP(1)', function (error, results, fields, queryTime) {
    console.log(queryTime); // 1.005
});

```

I didn't use a new option to handle this. That's because it doesn't have the overhead that is serious enough to worry about.

I hope this helps.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -281,6 +281,26 @@ Connection.prototype.format = function(sql, values) {
   return SqlString.format(sql, values, this.config.stringifyObjects, this.config.timezone);
 };
 
+Connection.prototype.enableQueryTimeProfiling = function(callback) {
+  this._isQueryTimeProfilingMode = true;
+
+  if (callback) {
+    this.on('queryTimeProfiling', function(query) {
+      callback(query);
+    });
+  }
+};
+
+Connection.prototype.disableQueryTimeProfiling = function() {
+  this._isQueryTimeProfilingMode = false;
+
+  this.removeAllListeners('queryTimeProfiling');
+};
+
+Connection.prototype.isQueryTimeProfilingMode = function() {
+  return this._isQueryTimeProfilingMode || false;
+};
+
 if (tls.TLSSocket) {
   // 0.11+ environment
   Connection.prototype._startTLS = function _startTLS(onSecure) {
--- a/lib/protocol/sequences/Query.js
+++ b/lib/protocol/sequences/Query.js
@@ -26,6 +26,7 @@ function Query(options, callback) {
 }
 
 Query.prototype.start = function() {
+  this._saveQueryTime('request');
   this.emit('packet', new Packets.ComQueryPacket(this.sql));
 };
 
@@ -89,6 +90,8 @@ Query.prototype['ErrorPacket'] = function(packet) {
 };
 
 Query.prototype['ResultSetHeaderPacket'] = function(packet) {
+  this._saveQueryTime('first_response');
+
   if (packet.fieldCount === null) {
     this._sendLocalDataFile(packet.extra);
   } else {
@@ -103,8 +106,12 @@ Query.prototype['FieldPacket'] = function(packet) {
 Query.prototype['EofPacket'] = function(packet) {
   this._resultSet.eofPackets.push(packet);
 
-  if (this._resultSet.eofPackets.length === 1 && !this._callback) {
-    this.emit('fields', this._resultSet.fieldPackets, this._index);
+  if (this._resultSet.eofPackets.length === 1) {
+    this._saveQueryTime('fields');
+
+    if (!this._callback) {
+      this.emit('fields', this._resultSet.fieldPackets, this._index);
+    }
   }
 
   if (this._resultSet.eofPackets.length !== 2) {
@@ -126,6 +133,8 @@ Query.prototype._handleFinalResultPacket = function(packet) {
     return;
   }
 
+  this._saveQueryTime('end');
+
   var results = (this._results.length > 1)
     ? this._results
     : this._results[0];
@@ -138,6 +147,11 @@ Query.prototype._handleFinalResultPacket = function(packet) {
 };
 
 Query.prototype['RowDataPacket'] = function(packet, parser, connection) {
+  if (typeof this._hasRows === 'undefined') {
+    this._saveQueryTime('first_row');
+    this._hasRows = true;
+  }
+
   packet.parse(parser, this._resultSet.fieldPackets, this.typeCast, this.nestTables, connection);
 
   if (this._callback) {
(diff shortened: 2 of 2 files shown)


dougwilson:
Is there no way to get the actual execution time on the server, exclusive of the network transfer and latency, like how the mysql command client works?

ifsnow:
@dougwilson I'll try to check how the mysql command client works.

dougwilson:
👍 I'm not 100% sure it does anything different, but I think I looked a long time ago and saw some kind of way it could better get the query time. The biggest issue with doing it in Node.js space is that it can get really off when the event loop is saturated or some tasking is blocking up the loop, ignoring the network latency issue.

sidorares:
I'm probably -1 to the way it is (just measuring before/after query time) as opposed to server side timing (if that's possible at all)

you can do this manually quite easily: 

```js
var start = new Date();
connection.query('SELECT * FROM `books` WHERE `author` = "David"', function (error, results, fields) {
  var queryTime = new Date() - start
})
```

sidorares:
also, extra parameter can potentially break old code (so this should be at least major semver) - see example here - https://github.com/sidorares/node-mysql2/pull/45#issuecomment-24448470

ifsnow:
@sidorares I totally agree on your opinion, but I thought it would be useful to provide it by default.

Thanks for letting me know the problem of using a extra parameter. I'll try to find the another approach.

I hope there's a good way.

ifsnow:
I took a couple of comments and changed my previous work.

@dougwilson mysql command line is also very similar.

https://github.com/mysql/mysql-server/blob/71f48ab393bce80a59e5a2e498cd1f46f6b43f9a/client/mysql.cc#L3613
(#) start_timer -> mysql_real_query -> mysql_store_result -> (#) end_timer -> mysql_fetch_row

I adjusted the position to end the timer.

@sidorares removed the additional parameter and chose a different approach.

```js
connection.query('SELECT * FROM `books` WHERE `author` = "David"', function (error, results, fields) {
  console.log(connection.getLastQueryTime());
});
```

sidorares:
@ifsnow maybe better to use `process.hrtime()` for sub-millisecond queries? ( I can see that happen on local connection )

sidorares:
Also might be interesting to log more then just a 'last row' time, but also TTFB / fields / first row / last row

ifsnow:
@sidorares process.hrtime() added in v0.7.6. This module supports version 0.6 or later, so it's likely to be difficult to use it. I thought about what you said (TTFB / fields / first row / last row), but I wonder if most of people want to know about that generally. If yes, it's not difficult.

@dougwilson What do you think about this?

sidorares:
@ifsnow my point is that just timing is vert easy to implement on top of public api if somebody needs  but more detailed breakdown need library support

> This module supports version 0.6 or later

fall back to Date() in that case?

ifsnow:
@sidorares I know what you're talking about, and I agree on offering detailed information if possible. I'll work on it a bit later.

ifsnow:
@sidorares I tried to improve this by a new approach. This new version can get detailed step-by-step times and uses process.hrtime. (included a fallback for nodejs v0.6)

```js
connection.enableQueryTimeProfiling(); // You can turn it on and off.

connection.query('SELECT * FROM `books` WHERE `author` = "David"', function(err, results, fields) { 
  connection.printQueryTimeProfiling(); // or connection.getQueryTimeProfiling();

  /*
  output example
  
  ================================
  - first_byte : 0.000886171 sec.
  - fields     : 0.000983313 sec.
  - first_row  : 0.000086899 sec.
  - last_row   : 0.000544853 sec.

  # total time : 0.002501236 sec.
  ================================  
  */
});
```

What do you think about this try?

dougwilson:
@ifsnow nice! Do you think the recording methods should be on the Query object instead of the Connection object? I'm thinking the Query object may make more sense since what you're trying to measure is the Query sequence time, not just all the packets on a given connection. It also seems like that would play really nice with the fact that you can have multiple queries all queued up on a single connection.

Also, regarding the output to console, do you know if that would work for the user you're implementing this for?

ifsnow:
@dougwilson It seems to be a good opinion, so I had an another try to be able to use this more easier.

```js
connection.enableQueryTimeProfiling(function (query) {
  query.printQueryTimeProfiling();
  
  /* output sample  
  ===============================================================================
  SELECT * FROM `books` WHERE `author` = "David"
  -------------------------------------------------------------------------------
  - first_byte : 0.000912604 sec.
  - fields     : 0.000930777 sec.
  - first_row  : 0.000084256 sec.
  - last_row   : 0.000513133 sec.

  # total time : 0.00244077 sec.
  ===============================================================================
  */
  
  query.getQueryTimeProfiling();
  
  /* result sample
  {
    sql: 'SELECT * FROM mysql.help_topic LIMIT 2',
    timeline: {
      first_byte: 0.00084553,
      fields: 0.000905997,
      first_row: 0.000077647,
      last_row: 0.000513133 
    },
    total_time: '0.002342307'
   }
   */
});

connection.query('SELECT * FROM `books` WHERE `author` = "David"', function(err, results, fields) { 
  ....
});
```

This approach doesn't add code for time checking to existing logic. I think this is a good direction to make the clean code, and output added a executed query to understand easily.


```js
connection.enableQueryTimeProfiling();

connection.query('SELECT * FROM `books` WHERE `author` = "David"', function(err, results, fields) { 
  this.printQueryTimeProfiling(); // or this.getQueryTimeProfiling();
  ....
});
```

This method is also offer in callback.

**If you agree on this, I'll add a description about using this at Readme.md**

dougwilson:
I guess what I meant above is that I don't think you need any methods on the Connection object at all, since things like conn.query() return the Query object. Even then, since conn.query() takes an object of options, it may make sense to just add the opt-in to that set of options instead of a bunch of new toggling methods on Connection. Unless I'm missing something here.

ifsnow:
@dougwilson I thought of using options of query, but it's so inconvinient to use to existing codes. That's because the code shoud be changed to an object of options. If there are lots of queries, It's not easy.

I don't like to change the code for testing whenever I need it. I think there is a possibility of a mistake in this process. People including me easily forget. :)

There is also the advantage of processing separately from the query. Now this approach can easily monitor all queries and filter out slow queries without affecting existing codes.

oxygen:
I just wrote this today to enable building performance counters on queries.
https://github.com/oxygen/unparametrize-sql-query

It is suposed to unparametrize queries the way the MySQL Performance Schema does.

Has the query execution time feature made it, or will it ever?

oxygen:
FYI

The old way of doing query profiling, using `SET SESSION profiling = 1;` is deprecated: 

"_19:29:26 SET SESSION profiling = 10 row(s) affected, 1 warning(s): 1287 '@@ profiling' is deprecated and will be removed in a future release.	0.031 sec`_"

https://dev.mysql.com/doc/refman/5.6/en/show-profile.html

The new way of doing profiling is using [Query Profiling Using Performance Schema](https://dev.mysql.com/doc/refman/5.6/en/performance-schema-query-profiling.html)
which is permanently disabled for many cloud providers.

**Which means, asking @ifsnow to provide data extracted from MySQL will not work for many people out of the box. It will also not work for production systems with performance schema instrumentation disabled.**

Issue #962, opened 2014-12-18
It would be good to add some idleTimeout to pool options


dusmanija:
Hello,

I am using this module for a long time but it lacks one failsafe element. My code for database functions became very big and it is hard to find where connection is not closed or if some errors happened before connection is closed that connection will stay idle and eventually when all are consumed from the pool the application hangs on database calls. It would be good to add some option like idelConnectionTimeout that will destroy a connection from a pool if it is idle for set time. This way pool will never stuck and will create fresh connections to replace those that hanged. Also if some mechanism could be made to auto close connection on end that would be great but i think that is not an option...

seme1:
Same here. Please implement this feature.

lroal:
This would be a nice feature.

dougwilson:
PRs are welcome :)!

brendanashworth:
Doesn't #802 do this?

dougwilson:
No

Nepherius:
This wound a good.

AdriVanHoudt:
+1 for this, now to find time for a PR :smiley:

akuma:
+1

PR #1749, opened 2017-06-06
Remove Stale connections that have not been used for idletimeout.

aankur:
#962 Add ability to remove connections from the pool which have been stale for idletimeout period in milliseconds

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -17,6 +17,39 @@ function Pool(options) {
   this._freeConnections      = [];
   this._connectionQueue      = [];
   this._closed               = false;
+  this._recycleTimer         = null;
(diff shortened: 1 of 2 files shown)


4ver:
Might be a good idea to change the following `shift` to `pop` to prevent the pool from cycling through the connections keeping em all alive.  

https://github.com/mysqljs/mysql/blob/master/lib/Pool.js#L37

adriancooney:
Any more movement on this? Our instances are hogging connections for too long and it'd be great to be able to close out those idle connections.

akleiber:
duplicate of https://github.com/mysqljs/mysql/pull/2218

PR #1776, opened 2017-07-20
Support authentication switch request and basic auth plugin. 

### LINKED ISSUE #1396: Node-mysql does not support auth-switch request from server.

elemount:
The new code supports mysql_native_password and mysql_old_password as potential authentication methods.
And for mysql_old_password, will still need `{inscure:true}`.
It  fixed #1396 and also fixed authentication against Azure Database for MySQL: #1729 .

The implementation refer the pull request #1730 .

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -112,7 +112,9 @@ Pool.prototype.acquireConnection = function acquireConnection(connection, cb) {
 
   if (changeUser) {
     // restore user back to pool configuration
+    var previousClientPluginAuth = connection.config.clientPluginAuth;
     connection.config = this.config.newConnectionConfig();
+    connection.config.clientPluginAuth = previousClientPluginAuth;
     connection.changeUser({timeout: this.config.acquireTimeout}, onOperationComplete);
   } else {
     // ping connection
--- a/lib/protocol/Auth.js
+++ b/lib/protocol/Auth.js
@@ -159,6 +159,8 @@ Auth.tokenByPlugin = function(pluginName, pluginData, password) {
     case 'mysql_old_password':
       // mysql_old_password only need first 8 bytes scramble.
       return this.scramble323(pluginData.slice(0, 8), password);
+    case 'mysql_clear_password':
+      return password;
     default:
       var err   = new Error('The auth plugin: ' + pluginName + ' is not supported by this client driver.');
       err.code  = 'UNSUPPORTED_AUTH_PLUGIN';
--- a/lib/protocol/Auth.js
+++ b/lib/protocol/Auth.js
@@ -150,3 +150,19 @@ Auth.int32Read = function(buffer, offset){
        + (buffer[offset + 2] << 8)
        + (buffer[offset + 3]);
 };
+
+Auth.tokenByPlugin = function(pluginName, pluginData, password) {
+  switch (pluginName) {
+    case 'mysql_native_password':
+      // mysql_native_password only need first 20 bytes scramble.
+      return this.token(password, pluginData.slice(0, 20));
+    case 'mysql_old_password':
+      // mysql_old_password only need first 8 bytes scramble.
+      return this.scramble323(pluginData.slice(0, 8), password);
+    default:
+      var err   = new Error('The auth plugin: ' + pluginName + ' is not supported by this client driver.');
+      err.code  = 'UNSUPPORTED_AUTH_PLUGIN';
+      err.fatal = true;
+      throw err;
+  }
+};
--- a/lib/protocol/packets/AuthSwitchPacket.js
+++ b/lib/protocol/packets/AuthSwitchPacket.js
@@ -1,0 +1,13 @@
+module.exports = AuthSwitchPacket;
+function AuthSwitchPacket(options) {
+  options = options || {};
+  this.scrambleBuff = options.scrambleBuff;
+}
+
+AuthSwitchPacket.prototype.parse = function(parser) {
+  this.scrambleBuff = parser.parsePacketTerminatedBuffer();
+};
+
+AuthSwitchPacket.prototype.write = function(writer) {
+  writer.writeBuffer(this.scrambleBuff);
+};
--- a/lib/protocol/packets/ComChangeUserPacket.js
+++ b/lib/protocol/packets/ComChangeUserPacket.js
@@ -2,11 +2,13 @@ module.exports = ComChangeUserPacket;
 function ComChangeUserPacket(options) {
   options = options || {};
 
-  this.command       = 0x11;
-  this.user          = options.user;
-  this.scrambleBuff  = options.scrambleBuff;
-  this.database      = options.database;
-  this.charsetNumber = options.charsetNumber;
+  this.command          = 0x11;
+  this.user             = options.user;
+  this.scrambleBuff     = options.scrambleBuff;
+  this.database         = options.database;
+  this.charsetNumber    = options.charsetNumber;
+  this.clientPluginAuth = options.clientPluginAuth;
+  this.authPluginName   = options.authPluginName;
 }
 
 ComChangeUserPacket.prototype.parse = function(parser) {
@@ -14,7 +16,12 @@ ComChangeUserPacket.prototype.parse = function(parser) {
   this.user          = parser.parseNullTerminatedString();
   this.scrambleBuff  = parser.parseLengthCodedBuffer();
   this.database      = parser.parseNullTerminatedString();
-  this.charsetNumber = parser.parseUnsignedNumber(1);
+  if (!parser.reachedPacketEnd()) {
+    this.charsetNumber = parser.parseUnsignedNumber(2);
+    if (this.clientPluginAuth === true) {
+      this.authPluginName = parser.parseNullTerminatedString();
+    }
+  }
 };
 
 ComChangeUserPacket.prototype.write = function(writer) {
@@ -23,4 +30,7 @@ ComChangeUserPacket.prototype.write = function(writer) {
   writer.writeLengthCodedBuffer(this.scrambleBuff);
   writer.writeNullTerminatedString(this.database);
   writer.writeUnsignedNumber(2, this.charsetNumber);
+  if (this.clientPluginAuth) {
+    writer.writeNullTerminatedString(this.authPluginName);
+  }
 };
--- a/lib/protocol/packets/HandshakeInitializationPacket.js
+++ b/lib/protocol/packets/HandshakeInitializationPacket.js
@@ -20,10 +20,16 @@ function HandshakeInitializationPacket(options) {
   this.filler3             = options.filler3;
   this.pluginData          = options.pluginData;
   this.protocol41          = options.protocol41;
+  this.clientPluginAuth    = options.clientPluginAuth;
+  this.authPluginName      = options.authPluginName;
 
   if (this.protocol41) {
     // force set the bit in serverCapabilities1
     this.serverCapabilities1 |= Client.CLIENT_PROTOCOL_41;
+  }
+
+  if (this.clientPluginAuth) {
+    this.serverCapabilities2 |= Client.CLIENT_PLUGIN_AUTH >> 16;
   }
 }
 
@@ -38,16 +44,22 @@ HandshakeInitializationPacket.prototype.parse = function(parser) {
   this.serverStatus        = parser.parseUnsignedNumber(2);
 
   this.protocol41          = (this.serverCapabilities1 & (1 << 9)) > 0;
+  this.clientPluginAuth    = false;
 
   if (this.protocol41) {
     this.serverCapabilities2 = parser.parseUnsignedNumber(2);
+    this.clientPluginAuth    = (this.serverCapabilities2 & (1 << 3)) > 0;
     this.scrambleLength      = parser.parseUnsignedNumber(1);
     this.filler2             = parser.parseFiller(10);
     // scrambleBuff2 should be 0x00 terminated, but sphinx does not do this
     // so we assume scrambleBuff2 to be 12 byte and treat the next byte as a
     // filler byte.
-    this.scrambleBuff2       = parser.parseBuffer(12);
+    scrambleBuff2Length      = Math.max(12, this.scrambleLength - 9);
+    this.scrambleBuff2       = parser.parseBuffer(scrambleBuff2Length);
     this.filler3             = parser.parseFiller(1);
+    if (this.clientPluginAuth) {
+      this.authPluginName    = parser.parseNullTerminatedString();
+    }
   } else {
     this.filler2             = parser.parseFiller(13);
   }
@@ -80,8 +92,11 @@ HandshakeInitializationPacket.prototype.write = function(writer) {
     writer.writeUnsignedNumber(2, this.serverCapabilities2);
     writer.writeUnsignedNumber(1, this.scrambleLength);
     writer.writeFiller(10);
+    writer.writeNullTerminatedBuffer(this.scrambleBuff2);
+    if (this.clientPluginAuth) {
+      writer.writeNullTerminatedString(this.authPluginName);
+    }
   }
-  writer.writeNullTerminatedBuffer(this.scrambleBuff2);
 
   if (this.pluginData !== undefined) {
     writer.writeNullTerminatedString(this.pluginData);
--- a/lib/protocol/sequences/ChangeUser.js
+++ b/lib/protocol/sequences/ChangeUser.js
@@ -8,11 +8,12 @@ Util.inherits(ChangeUser, Sequence);
 function ChangeUser(options, callback) {
   Sequence.call(this, options, callback);
 
-  this._user          = options.user;
-  this._password      = options.password;
-  this._database      = options.database;
-  this._charsetNumber = options.charsetNumber;
-  this._currentConfig = options.currentConfig;
+  this._user           = options.user;
+  this._password       = options.password;
+  this._database       = options.database;
+  this._charsetNumber  = options.charsetNumber;
+  this._currentConfig  = options.currentConfig;
+  this._authPluginName = options.authPluginName;
 }
 
 ChangeUser.prototype.start = function(handshakeInitializationPacket) {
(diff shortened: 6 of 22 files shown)


lib/protocol/Auth.js
```diff
@@ -150,3 +150,16 @@ Auth.int32Read = function(buffer, offset){
...
+Auth.tokenByPlugin = function(pluginName, pluginData, password) {
+  switch (pluginName) {
+    case 'mysql_native_password':
+      // mysql_native_password only need first 20 bytes scremble.
+      return this.token(password, pluginData.slice(0, 20));
+    case 'mysql_old_password':
+      // mysql_old_password only need first 8 bytes scremble.
+      return this.scramble323(pluginData.slice(0, 8), password);
+    default:
+      throw 'Unknown auth plugin: ' + pluginName;
```
dougwilson: This should throw an error object, with a proper code and fatal flag so users can handle it like the other errors.

elemount: Sure.

dougwilson:
Awesome, thanks @elemount ! The last question I had here is that I believe you are the one who asked about sha256_password authentication plugin support. This pull request doesn't add that, but does this pull request implement the plugin system talked about so you'd be able to code sha256_password authentication plugin support in your code? If so, there's no documentation about how one would go about this in the pull request. If not, what is the plan to add sha256_password authentication plugin (or any custom plugin users may have setup in their MySQL)? Just curious because once this lands, if adding plugin support on top of this is hard, probably going to be very plainful to land when we already know user plugins is a use-case.

elemount:
@dougwilson , you raise a great problem for me. 

In fact, I still want to add sha256_password_plugin. But with some thought, I think put the sha256_password_plugin in this pull request is not wise. 

The main problem is the tests. In this pull request, all the code is both in UT and integration test. But for sha256_password_plugin, it hard to use integration tests. I have a local MySQL environment to test the sha256_password_pluign, to make other one test this plugin, I try to found a MySQL docker image with openssl build support, which can fully test the sha256_password_plugin, but I failed. Now I decide to build such docker image by myself, and can deliver to other people to test it easily. But it also needs some time. 

Another one is the design. I see both [Golang driver](https://github.com/go-sql-driver/mysql) and Python driver [PyMySQL](https://github.com/PyMySQL/PyMySQL) has a pluggable auth plugin like that [pull request](https://github.com/go-sql-driver/mysql/pull/552), I think this pluggable auth plugin is a good design for that. 

In my plan, I want to
1. Support basic client plugin auth attribute in this pull request.
2. Support pluggable auth plugin in other pull request including make auth plugin can both provide by users or by other contributors. To make add auth plugin easily, in my mind, contributors should only just modify one definition file and just add a new file. And user can provide a class with a name to add their personal auth plugin. 
3. Add sha256_password_plugin by the pluggable auth plugin mechanism.

The pluggable auth plugin must be supported in future because I see MariaDB is planning to have another plugin https://mariadb.org/history-of-mysql-mariadb-authentication-protocols/ . And if user want to use Amazon RDS with IAM plugin, they can also just provide a class to the connection string. 

But I do not want to merge the client plugin auth attribute and the pluggable auth plugin in one pull request. The main problem is complex, and I also concern that you may not accept my pluggable auth plugin design. The client plugin auth attribute is the first step, but also the basic step. Mark the attribute ON firstly and then in next pull request we have a pluggable auth plugin and then sha256_password_plugin and other auth plugins, step by step.

Give me your comments on my plan, or what's your plan on it?

sidorares:
@elemount @dougwilson I am going to change auth plugin api to make it easier to configure especially if you need to handle multiple plugins - https://github.com/sidorares/node-mysql2/issues/560#issuecomment-312151878

Current api documented here: https://github.com/sidorares/node-mysql2/blob/master/documentation/Authentication-Switch.md

```js
const conn = mysql.createClient({
  authPlugins: {
    some_custom_plugin: (data, cb) => { /* ... * /}
    // mysql_native_plugin and sha256_password_plugin automaticlly merged to here unless you do something like
    // mysql_native_plugin: null
    // 'mysql_clear_password' not enabled by default:
    mysql_clear_password: mysql.authPlugins.mysql_clear_password
  },
  connectAuthPluginName: 'mysql_native_password' // this woukd be default, but if you need to start with sha256_password_plugin for example you need to put this option
})
```
would be good to add AD and PAM plugins to examples but no need to bundle them with driver

elemount:
@sidorares I think this design is great. 
And @dougwilson , what's your options on the pluggable auth plugin design?
Do you think that I should implement this pluggable auth plugin in this pull request?

sidorares:
@elemount note that in theory there might me multiple data roundtrips between server and auth plugins, not just single helloFromServer(plugin name, hello data) => helloFromClient( client hello response data ) . Maybe there should be extra parameter?  `(data, cb, sequenceId) => {}` ? (though this can be solved using closure counter )

elemount:
@sidorares , In fact, I do not think `({pluginName, pluginData, sequenceId}, cb)` is better than `({pluginName, pluginData}, cb)`.
For example, in `sha256_password_plugin`, this plugin must know which phase it locates, and the phase is not only depends on the sequenceId, it depends on the previous state of the auth plugin (for example, received the public_key can be on sequenceId 1 or 2).
So I think sequenceId can simplify some auth plugin, but it is not a common solution.

In my mind, if we describe the connection as a state machine, the pluggable auth plugin is a node which we impose to contributors and users to implement. If the auth plugin is just a node(a function), for complicated auth plugin, it must be a node of stack based state machine, which means it can pop its last push state and push its newest state. SequenceId is means that the when function is called, then it pop a id, and after it finished, it push the next id, not flexible enough, and also not usable enough.
 
I advise you can describe how to implement complicated auth plugin in the document sample, instead of the uncompleted solution `sequenceId` for just simplify a few plugins.

sidorares:
```js
 const createPlugin = () => {
  let state = 'start';
  return (data, cb) => {
    if (state === 'start') {
       state = 'start2';
       return cb(null, 'foo');
    }
    if (state === 'start2') {
       return cb(null, 'bar');
    }
  });

  const conn = mysql.createClient({
    authPlugins: {
      foo_bar_plugin = createPlugin()
    }
  })
```

elemount:
Or what about that, @sidorares ?
```javascript
function ({pluginName, pluginData, extraPluginInfo}, cb) {
    if (pluginName === 'ssh-key-auth') {
      getPrivateKey(function (key) {
        var response = encrypt(key, pluginData);
        if (extraPluginInfo == 'phase1') {
          return cb(null, response, 'phase2' /*this is the new extraPluginInfo*/);
        } else {
          // continue handshake by sending response data
          // respond with error to propagate error to connect/changeUser handlers
          cb(null, response, 'end')
        };
      });
    } else {
      const err = new Error(`Unknown AuthSwitchRequest plugin name ${pluginName}`);
      err.fatal = true;
      cb(err);
    }
```

elemount:
@sidorares Could you help to review the code first?

sidorares:
looks good to me. Re parity with mysql2 - this pr currently does not add user-supplied pluggable auth and only implements standard mysql_native_password and mysql_old_password plugins, so no new api at the moment

sidorares:
@elemount can you rebase pr please?

elemount:
@sidorares I've rebased it.

dougwilson:
@elemount does this PR take into account all the comments to PR #1730 you linked to? For example, one thing we talked about was adding an integration test so we can know if at some point the string[EOF] vs string[NUL] bug is fixed in MySQL, but I'm not sure I see that in this PR, but maybe I'm missing it? Can you point me to it?

lib/protocol/packets/ComChangeUserPacket.js
```diff
@@ -2,19 +2,25 @@ module.exports = ComChangeUserPacket;
...
 }
 
 ComChangeUserPacket.prototype.parse = function(parser) {
   this.command       = parser.parseUnsignedNumber(1);
   this.user          = parser.parseNullTerminatedString();
   this.scrambleBuff  = parser.parseLengthCodedBuffer();
   this.database      = parser.parseNullTerminatedString();
-  this.charsetNumber = parser.parseUnsignedNumber(1);
+  this.charsetNumber = parser.parseUnsignedNumber(2);
+  if (!parser.reachedPacketEnd()) {
```
dougwilson: Shouldn't this key off the fact that plugin auth is used, not if packet end was reached? For example, this will not work with CLIENT_CONNECT_ATTRS because if client auth is off but CLIENT_CONNECT_ATTRS is on, then the packet end is indeed not reached, but the rest of the packet is not related to client plugin auth.

dougwilson: Docs http://web.archive.org/web/20160604101506/http://dev.mysql.com/doc/internals/en/com-change-user.html

dougwilson:
And forgive me if you answered my question, but even re-reading your responses I'm still not sure: what is the plan to add sha256_password authentication plugin (or any custom plugin users may have setup in their MySQL)? Just curious because once this lands, if adding plugin support on top of this is results in a breaking change, it may end up never landing...

dougwilson:
Also, forgot to note: please add some information in the README about this, for example the list of plugins supported. Also update the https://github.com/mysqljs/mysql#connection-flags section with the changes you made to the default flags.

elemount:
@dougwilson , I use another way to **avoid** the string[EOF] and string[NUL] bug problem.

In protocol level, the AuthSwitch packet contains `'auth_plugin_name\x00auth_plugin_data'`, `'auth_plugin_name\x00'`  is a string[NUL] and the `'auth_plugin_data'` is string[EOF].

In real world level, the `'auth_plugin_data'` which 'mysql_native_password' is 20bit scramble and another '\x00' which is 21bit, that is a string[NUL], the 'mysql_old_password' is 8bit scramble and another '\x00' is 9bit, that is not match the protocol.

But fortunately we do not need to handle this mismatch. How? Also by the protocol.
I follow the protocol `'auth_plugin_name\x00'`  is a string[NUL] and the `'auth_plugin_data'` is string[EOF]. The '\x00' ends the auth_plugin_data, but we do not need to care about it.
MySQL describe the what mysql_native_password and mysql_old_password need to accept.
In the page [Old Password Authentication](http://web.archive.org/web/20170404224445/https://dev.mysql.com/doc/internals/en/old-password-authentication.html) and [Native Password Authentication](http://web.archive.org/web/20170405005641/https://dev.mysql.com/doc/internals/en/secure-password-authentication.html), mysql_old_password needs first 8 byte of `auth_plugin_data` and mysql_native_password needs first 20 byte of `auth_plugin_data`. We do not need care the 9th bit and the 21th bit, we do not need care its value, even it is '\x00'.

Thus I avoid the real world mismatch by following the MySQL client protocol.

dougwilson:
Right, I'm fine with that slice, but what I'm talking about is the underlying packet read. Because the protocol and the MySQL server implementation disagree, this is an important aspect to make sure we have an actual integration test, and not simply a unit test. This is important when we test against other implementations like MarisDB, Azure, and any others that popup (like various MySQL proxy software) to be aware when we run aground.

lib/protocol/packets/ComChangeUserPacket.js
elemount: Sure. Let me change the parse.

elemount:
@dougwilson , I will create PR on pluggable auth plugin and sha256_password auth support in this month, I have enough time to finished it this month and I'll do it, at least have a pull request with ut and integration test. 

And also I can make sure it will not have a breaking change which introduces the pluggable auth plugin.

elemount:
@dougwilson , I do not see any underlying packet read problem.
In my mind, which may be wrong. The protocol is `string[EOF]`, so we just readPacketToEnd, readPacketToEnd is the only correct one because it follows the protocol. And we slice to get the scramble, then we do auth.  No packet read problem, the protocol describe a `string[EOF]`, then why we need to care about the content of `string[EOF]`, do we need to know it there any `\x00` in the string? The sramble can be `\0\0....\0`, but it is not a series of `string[NUL]`s, it just a `string[EOF]`, we follow it and any other server also aspect it. Then can add `\0` in the end of the `string[EOF]`, but not breaking anyone.

What your insight?

elemount:
Hi @dougwilson , @sidorares , I resolved all the comments, please help to review it again.

twocode:
Great to this pr. I created an issue on this a while ago and have been using mysql2 ever since thanks to @sidorares supported it. I am looking forward to see this pr gets merged. Great job, @elemount !

lib/protocol/packets/ComChangeUserPacket.js
```diff
@@ -2,19 +2,26 @@ module.exports = ComChangeUserPacket;
...
+  this.authPluginName   = options.authPluginName;
 }
 
 ComChangeUserPacket.prototype.parse = function(parser) {
   this.command       = parser.parseUnsignedNumber(1);
   this.user          = parser.parseNullTerminatedString();
   this.scrambleBuff  = parser.parseLengthCodedBuffer();
   this.database      = parser.parseNullTerminatedString();
-  this.charsetNumber = parser.parseUnsignedNumber(1);
+  if (!parser.reachedPacketEnd()) {
```
dougwilson: Why this check here? Shouldn't the charsetNumber just be outside the if like it used to be, or is there a reason behind this change?

lib/protocol/sequences/Handshake.js
```diff
@@ -104,3 +122,14 @@ Handshake.prototype['ErrorPacket'] = function(packet) {
   err.fatal = true;
   this.end(err);
 };
+
+function HandshakeInsecureAuthError() {
+  var err   = new Error(
+    'MySQL server is requesting the old and insecure pre-4.1 auth mechanism.' +
```
dougwilson: Looks like the space that was at the end of this sentence got removed.

lib/protocol/packets/ComChangeUserPacket.js
elemount: Hi @dougwilson , this is for the document [COM_CHANGE_USER packet](https://web.archive.org/web/20170404163330/https://dev.mysql.com/doc/internals/en/com-change-user.html), it include the charset in the `if more data` condition.

lib/protocol/sequences/Handshake.js
elemount: Sure. Fixed it.

lib/protocol/packets/ComChangeUserPacket.js
dougwilson: Oh, sorry :) I was thinking the handshake. Nevermind on this, then :)

elemount:
@dougwilson , I've fixed the README issue and rebased the code. Could you merge this code after check?

lib/protocol/sequences/ChangeUser.js
```diff
@@ -34,8 +37,31 @@ ChangeUser.prototype.start = function(handshakeInitializationPacket) {
   this.emit('packet', packet);
 };
 
+ChangeUser.prototype.determinePacket = function(firstByte) {
+  if (firstByte === 0xff) {
+    return Packets.ErrorPacket;
+  }
+
+  if (firstByte === 0xfe) {
```
dougwilson: I am fixing up this edge case where the server can actually respond here with the old password packet, of 0xfe doesn't always mean auth switch as implemented here, so just gotta add the same logic from the handshake.

dougwilson:
I see you pushed again. I'm going to comment on the places where I am making changes to help show that I'm busy working on this. If you're making changes too, let me know so I can stop for now while you make changes...

lib/protocol/packets/AuthSwitchPacket.js
```diff
@@ -0,0 +1,13 @@
+module.exports = AuthSwitchPacket;
+function AuthSwitchPacket(options) {
+  options = options || {};
+  this.scrambleBuff = options.scrambleBuff;
```
dougwilson: Since this is greenfield, I'm just changing these props to have the same name as in the protocol docs for sanity.

lib/protocol/packets/ComChangeUserPacket.js
dougwilson: I'm writing a test to exercise this case, there the charsetNumber may not get populated.

lib/protocol/packets/AuthSwitchPacket.js
elemount: Sure.

lib/protocol/packets/HandshakeInitializationPacket.js
```diff
@@ -80,8 +92,11 @@ HandshakeInitializationPacket.prototype.write = function(writer) {
     writer.writeUnsignedNumber(2, this.serverCapabilities2);
     writer.writeUnsignedNumber(1, this.scrambleLength);
     writer.writeFiller(10);
+    writer.writeNullTerminatedBuffer(this.scrambleBuff2);
+    if (this.clientPluginAuth) {
+      writer.writeNullTerminatedString(this.authPluginName);
+    }
   }
-  writer.writeNullTerminatedBuffer(this.scrambleBuff2);
```
dougwilson: This was a miss in the first, and since it's fixed in here, I'm trying to write a test case for this bug.

lib/Pool.js
```diff
@@ -112,7 +112,9 @@ Pool.prototype.acquireConnection = function acquireConnection(connection, cb) {
 
   if (changeUser) {
     // restore user back to pool configuration
+    var previousClientPluginAuth = connection.config.clientPluginAuth;
     connection.config = this.config.newConnectionConfig();
+    connection.config.clientPluginAuth = previousClientPluginAuth;
```
dougwilson: I need to rework how this is tracked so it stops changing the hidden class of the config object.

dougwilson:
Hey @elemount it is getting very late in my time zone and I'm likely to head out soon. I pushed up the PR #1730 code to `master` and I had this PR rebased on top of it already, but haven't yet rebased again the latest squash you pushed up. I need to do that still with the changes I noted above. Running the tests on this project just really eats up a lot of time :S

elemount:
@dougwilson , I add a commit on `mysql_clear_password` support. But faced merged issue. What should I do, merge the current code, or just left it?

And the I'm working on the pluggable auth plugin, how should I give the pull request, based on this pull request or have an another which its code based on it? Or I should fork on another branch?

dougwilson:
> @dougwilson , I add a commit on mysql_clear_password support. But faced merged issue. What should I do, merge the current code, or just left it?

Wow, not sure how I missed this, but was just going through open PRs / issues tonight to see where they are at. The best would have been to rebase the PR. I'm working on doing that now, because I don't see anything wrong with this PR, just needs to be rebased now.

> And the I'm working on the pluggable auth plugin, how should I give the pull request, based on this pull request or have an another which its code based on it? Or I should fork on another branch?

So it would depending I guess on what is needed. Ideally it would be a separate PR either way. If it _needs_ what is in this PR, then yea, just branch from this PR and submit a PR. Yes, at first that RP will include _all_ changes since they are not in our `master`, but after this gets merged, it would only have the diff of the two. Adding it to this PR would just stretch out landing this PR, which is the downside to mixing.

rhummelmose:
So is this actually fixed in the newest version?

PR #1805, opened 2017-08-18
Add stackCycle feature for decrease useless connections

davidshimjs:
If you are using a connection pool and suddenly handle more traffic than usual, the size of the connection pool increases. And the increased connections will survive because of the wait_timeout setting. (It usually set as 10-30 seconds) Keeping unnecessary connections is not good for running multiple servers.

However, this problem can not be solved because the mysql module only supports round-robin styles. So I added a pool option called stackCycle.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -369,7 +369,7 @@ time one is needed.
 Connections are lazily created by the pool. If you configure the pool to allow
 up to 100 connections, but only ever use 5 simultaneously, only 5 connections
 will be made. Connections are also cycled round-robin style, with connections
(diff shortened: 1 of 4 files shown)


davidshimjs:
@Robert-lihouyi What's your point? If you just want to listen answer, you can check out more information https://github.com/mysqljs/mysql#pooling-connections

Robert-lihouyi:
issue resolved, thanks very much

姓名:李厚毅
联系电话：18101082232
职位：高级前端工程师
________________________________

Issue #1803, opened 2017-08-15
pool.query fails after calling pool.end


HQidea:
Sometimes, I need to close the old pool each 5 mins and create a new one. So I can't make sure `pool.end` is called after all queries end. Since it is said 

> The end method takes an optional callback that you can use to know once all the connections have ended. The connections end gracefully, so all pending queries will still complete and the time to end the pool will vary.

in the doc, I think I can just call `pool.end` any time I want, but I will occasionally get an error: 

> Error: Pool is closed.

So,
What does *pending queries* mean in the doc?
Is a query which have been called `pool.query` a pending query?

For example, 
```js
var mysql = require('mysql');
var pool  = mysql.createPool({
  connectionLimit : 10,
  host            : 'example.org',
  user            : 'bob',
  password        : 'secret',
  database        : 'my_db'
});

pool.query('SELECT 1 + 1 AS solution', function (error, results, fields) {
  if (error) throw error;
  console.log('The solution is: ', results[0].solution);
});
pool.end();  // will be called somewhere else, but will raise the same error here
```

I've found this commit 76de66e2c36a6da71dcd010b3afe0029e5eb68bb fixed pool.getConnection race conditions, but I don't quite understand what are the race conditions here. In my opinion, after I have called `pool.query`, this query is a pending query no matter what phase the operation is under the hood such as ping.

elemount:
Hi @HQidea , the `pending query` is the query which has been dispatched on a free connection. Not all the queries which called `pool.query`. 

That is why your callback functions throw the error because the new queries can not find a connection for the pool is closed.

elemount:
@HQidea , could you close this issue?

HQidea:
Hi @elemount, does this count as an official response?

elemount:
Sorry, I'm not sure what's your problem? Could you clarify it?

HQidea:
To a user, there isn't any document contains the definition of `pending queries`. And since the pool will end gracefully, raising an error from previous query doesn't act as ending gracefully.

sidorares:
> raising an error from previous query doesn't act as ending gracefully.

"Error: Pool is closed." is not an error from one of the queries, it's "you are trying to close pool that is already closed" type of error. When you call `.end()` you should somehow remove your pool from 'active' state. For example:

```js
function refreshPool() {
   const poolToClose = pool;
   pool = null;
   if (poolToClose) {
     poolToClose.end();
   }
   pool = mysql.createPool(config);
}
```

[...]

PR #1810, opened 2017-08-28
feat: pool adds gracefulExit to end gracefully

HQidea:
As we discussed at #1803, this pull request adds an option to pool's creation,
```
var pool = common.createPool({
  // ...
  gracefulExit: true
});
```
If `gracefulExit` is `true`, every `pool.getConnection` or `pool.query` called before `pool.end` will success. If `false`, only commands / queries already in progress will complete, others will throw an error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 coverage/
 node_modules/
 npm-debug.log
+.DS_Store
--- a/Readme.md
+++ b/Readme.md
@@ -392,6 +392,7 @@ constructor. In addition to those options pools accept a few extras:
 * `queueLimit`: The maximum number of connection requests the pool will queue
   before returning an error from `getConnection`. If set to `0`, there is no
   limit to the number of queued connection requests. (Default: `0`)
+* `gracefulExit`: Determines whether to end gracefully. If `true`, every `pool.getConnection` or `pool.query` called before `pool.end` will success. If `false`, only commands / queries already in progress will complete, others will throw an error.
 
 ## Pool events
 
@@ -461,6 +462,11 @@ all the connections have ended.
 
 **Once `pool.end()` has been called, `pool.getConnection` and other operations
 can no longer be performed**
+
+If `gracefulExit` is set to `true`, the connections end _gracefully_, so all
+ -pending queries will still complete and the time to end the pool will vary.
+
+The default `gracefulExit` is `false`, the following behavior will take effect.
 
 This works by calling `connection.end()` on every active connection in the
 pool, which queues a `QUIT` packet on the connection. And sets a flag to
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -17,9 +17,10 @@ function Pool(options) {
   this._freeConnections      = [];
   this._connectionQueue      = [];
   this._closed               = false;
+  this._pendingClosing       = false;
 }
 
-Pool.prototype.getConnection = function (cb) {
+Pool.prototype.getConnection = function (cb, queued) {
 
   if (this._closed) {
     var err = new Error('Pool is closed.');
@@ -33,9 +34,18 @@ Pool.prototype.getConnection = function (cb) {
   var connection;
   var pool = this;
 
-  if (this._freeConnections.length > 0) {
+  if (this._freeConnections.length > 0 && (!this._pendingClosing || queued)) {
     connection = this._freeConnections.shift();
     this.acquireConnection(connection, cb);
+    return;
+  }
+
+  if (this._pendingClosing) {
+    var err = new Error('Pool is closed.');
+    err.code = 'POOL_CLOSED';
+    process.nextTick(function () {
+      cb(err);
+    });
     return;
   }
 
@@ -141,6 +151,10 @@ Pool.prototype.releaseConnection = function releaseConnection(connection) {
       this._freeConnections.push(connection);
       this.emit('release', connection);
     }
+
+    if (this._pendingClosing) {
+      this.end(this._endCallback);
+    }
   }
 
   if (this._closed) {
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 coverage/
 node_modules/
 npm-debug.log
+.DS_Store
```
dougwilson: You can remove this; setup a global ignore for your system if you need to ignore system-specific files.

Readme.md
```diff
@@ -392,6 +392,7 @@ constructor. In addition to those options pools accept a few extras:
 * `queueLimit`: The maximum number of connection requests the pool will queue
   before returning an error from `getConnection`. If set to `0`, there is no
   limit to the number of queued connection requests. (Default: `0`)
+* `gracefulExit`: Determines whether to end gracefully. If `true`, every `pool.getConnection` or `pool.query` called before `pool.end` will success. If `false`, only commands / queries already in progress will complete, others will throw an error.
```
dougwilson: Can the option be an option given to thr .end method instead of a pool option?

Readme.md
```diff
@@ -462,6 +463,11 @@ all the connections have ended.
 **Once `pool.end()` has been called, `pool.getConnection` and other operations
 can no longer be performed**
 
+If `gracefulExit` is set to `true`, the connections end _gracefully_, so all
+ -pending queries will still complete and the time to end the pool will vary.
+
+The default `gracefulExit` is `false`, the following behavior will take effect.
```
dougwilson: I think the exact details of how the graceful end works here would be useful, as I remember that was the initial confusion here. As far as I can tell, this implementation will still let checked out connections queue new queries if there is at least one connection in the middle of being established, since the QUIT commands are not queued on all the connections until there is no connection in the aquiring state, is that intentional?

lib/Pool.js
```diff
@@ -17,9 +17,10 @@ function Pool(options) {
   this._freeConnections      = [];
   this._connectionQueue      = [];
   this._closed               = false;
+  this._pendingClosing       = false;
 }
 
-Pool.prototype.getConnection = function (cb) {
+Pool.prototype.getConnection = function (cb, queued) {
```
dougwilson: Please add documentation for the new queued argument.

.gitignore
HQidea: ok

lib/Pool.js
HQidea: This queued argument is a private argument, is it appropriate to add documentation to README.md?

Readme.md
HQidea: Yes, it's intentional. I think graceful exit means finishing all queries called before end, so I make that happen.

Readme.md
HQidea: Yes

Readme.md
dougwilson: Right, but my comment was that your code is still sometimes allowing queries to be called after end. Is the call to end a hard stop, or should new queries still be allowed to get scheduled? The other issue is that under the acquiring condition, it will sometimes then stop the checked out connections from continuing to query and sometimes not.

lib/Pool.js
dougwilson: There is no such thing as a private argument -- the user can simply pass in the value. Just not documenting it doesn't make it private. If the argument is there, it needs to be documented.

lib/Pool.js
```diff
@@ -154,19 +168,22 @@ Pool.prototype.releaseConnection = function releaseConnection(connection) {
...
     // get connection with next waiting callback
-    this.getConnection(this._connectionQueue.shift());
+    this.getConnection(this._connectionQueue.shift(), true);
   }
 };
 
-Pool.prototype.end = function (cb) {
-  this._closed = true;
-
+Pool.prototype.end = function (gracefulExit, cb) {
```
dougwilson: So it can be expanded in the future, just make this an object instead of a plain Boolean. It'll match the signature of all the other methods that take options as well this way.

lib/Pool.js
```diff
@@ -177,14 +194,26 @@ Pool.prototype.end = function (cb) {
...
+    }
+
+    if (waitingClose === 0) {
+      process.nextTick(onEnd);
+    }
+    return;
   }
+
+  this._pendingClose = true;
+  this._endCallback = cb;
```
dougwilson: Please add this property into the constructor to avoid hidden class changes.

Readme.md
```diff
@@ -451,20 +451,37 @@ trying to gracefully shutdown a server. To end all the connections in the
...
 ```js
-pool.end(function (err) {
+pool.end(false, function (err) {
   // all connections in the pool have ended
 });
 ```
 
-The `end` method takes an _optional_ callback that you can use to know once
-all the connections have ended.
+The `end` method takes two _optional_ arguemnts:
```
dougwilson: typo: arguments

HQidea:
Sorry to fix late, and a test case as you requested was added.

Sprax2013:
I'd love to see this feature. Any progress on that?

Readme.md
HQidea: > Right, but my comment was that your code is still sometimes allowing queries to be called after end. Is the call to end a hard stop, or should new queries still be allowed to get scheduled?

Only queries have been called before end would be allowed to execute, new queries after end would not be allowed no matter whether there is any checked out connection in the pool

> The other issue is that under the acquiring condition, it will sometimes then stop the checked out connections from continuing to query and sometimes not.

When will the acquiring condition will happen?

HQidea:
@dougwilson I updated the Readme.md because of conflicts. So is there any chance to merge this pr?

dougwilson:
Hi @HQidea I really apologize on this one. I looked over the state of the PR when @Sprax2013 commented trying to remember why it wasn't merged. I got to work and found the reason in some notes I had: I meant to update the PR to pass options through the PoolCluster, which wraps around Pool so they could utilize the new end options added here.

I completely lost track of that, and I'm sorry for that.

HQidea:
Hi @dougwilson I added the option to PoolCluster.end. And should I add the option to PoolCluster.remove too?

kmaworld586:
unsubscribe this account

    On Friday, June 28, 2019, 4:55:06 AM CDT, HQidea <notifications@github.com> wrote:  
 
 
Hi @dougwilson I added the option to PoolCluster.end. And should I add the option to PoolCluster.remove too?

Ahmad45123:
Whats the progress on this ? Would love to see it in already :/ Is there an alternative for now ?

PR #1816, opened 2017-09-07
Ping on release to sync wait_timeout behind MaxSclae

E88:
Because Maxscale for example splits read and write queries to differemt cluster member, there are several connection from MaxScale to the DB server.
Each connections has its own timeout (wait_timeout) to be disconnected.
Example:
- wait_timeout 50 seconds
- 1 write query 
- 70 seconds read queries (write connection behind MaxScale never used, closed after 50s)
- 1 write query fail: "Conenction lost"

### DIFF
--- a/lib/Pool.js
+++ b/lib/Pool.js
@@ -17,6 +17,13 @@ function Pool(options) {
   this._freeConnections      = [];
   this._connectionQueue      = [];
   this._closed               = false;
+  
+  // Sync wait_timeout on connections behind MaxScale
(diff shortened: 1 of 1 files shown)


E88:
Better problem discriptin (wait_timeout is 10 seconds):
- getCon
- 1 write (query 'insert')
- 1 long time read query (query 'select sleep(20)') tresults in timeouts: read=0, write>10 closed
- ... read (select) works
- ... write (insert) fails: 'Connection Lost'

Issue #1887, opened 2017-11-28
connection.state not updating properly after closing with .end() or .release() methods.


alwaysautumn:
When calling `connection.end`  or `connection.release` with the optional callback, the value of `connection.state` will continue to be `'authenticated'`, though, if you try to run a query on the supposedly still open connection, an `Error: Cannot enqueue Query after invoking quit` will be thrown.

Here is the proof:

### Code snippet from _release-err-proof.js_

```
  // Check that the connection is open
  console.log('conn.state pre-release:', conn.state);
  
  // Attempt to close the connection gracefully
  conn.end((err) => {
    if (!err) {
       // Check that the connection was close properly
      console.log('conn.state post-release:', conn.state);

      conn.query('SELECT * FROM ??', [ 'Test' ] (err, results, fields) => {
        if (!err) {
          console.log('no error');
        } else {
          console.log('query err:', err);
        }
      });
    } else {
      console.log('release err:', err);
    }
  });
};

```

### Terminal output
```
$ node release-err-proof.js
> Attempting to close the connection...
> conn.state pre-release: authenticated
> conn.state post-release: authenticated
> query err: { Error: Cannot enqueue Query after invoking quit.
    at Protocol._validateEnqueue...
```

This behavior does not occur when the `connection.destroy` method is used; the value of `connection.state`, in this case, is `'disconnected'`.

This has been causing a lot of confusion, because I typically like to ensure that the connection has been properly closed via something like:

```
if (connection && connection.state === 'disconnected' {
  // Something something something javascript something something complete
} else {
  // Retry graceful connection close
}
```

Anyway, I am going to remove my `connection.state` check to alleviate the issue, but it would be nice to be able to ensure that the connections are actually closed. Please tell me if I am missing something and this feature is working as it should, otherwise, I'd love to see it fixed :D

Thanks!
Autumn

dougwilson:
I'll have to take a look, though `connection.state` is not a documented public API, as far as I know. This state is just for internal connection management and state changes, mostly around the state of the underlying TCP connection. In your example, the state hasn't moved to disconnected because when you check it, the TCP connection is still open at that point, thus why it's not "disconnected". Does that make sense? Since this is just an internal flag not documented, we of course don't have documentation around what the states mean (it's a private API).

alwaysautumn:
[...]

PR #1888, opened 2017-11-28
Added connection.state === 'disconnected' to Connection.prototype.end

alwaysautumn:
This was a pretty easy fix and should resolve the issue https://github.com/mysqljs/mysql/issues/1887 that I opened earlier. I did not make the changes to Pool.js, though. The same issues were cropping up with the `.release()` method.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -245,6 +245,7 @@ Connection.prototype.end = function end(options, callback) {
     opts.timeout = 30000;
   }
 
+  this.state = 'disconnected';
   this._implyConnect();
   this._protocol.quit(opts, bindToCurrentDomain(cb));
 };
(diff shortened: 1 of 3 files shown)


dougwilson:
The change here is causing the test suite to fail, it looks like, since the state is being set to disconnected prior to the TCP channel actually being closed.

dougwilson:
The connection management is still broken with the latest change -- the `state` should not be set to `disconnected` until the underlying TCP connection has closed. The change is setting it as soon as a sequence in the protocol has completed, which is too early.

dougwilson:
I'm going to close this since there hasn't been any update in a month. When you get a chance to fix it up, feel free to open a new PR 👍

PR #1926, opened 2018-01-23
Add a string template tag handler for securely composing queries.

mikesamuel:
This is a rough draft.  It is probably not suitable in its current
form.

https://nodesecroadmap.fyi/chapter-7/query-langs.html describes
this approach as part of a larger discussion about library support
for safe coding practices.

This enables

    connection.query`SELECT * FROM T WHERE x = ${x}, y = ${y}, z = ${z}`(callback)

and similar idioms.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,6 @@
 var Classes = Object.create(null);
+var calledAsTemplateTagQuick = require('template-tag-common')
+    .calledAsTemplateTagQuick;
 
 /**
  * Create a new Connection instance.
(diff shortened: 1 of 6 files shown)


mikesamuel:
It looks like tests pass for node >= 6 and fail for node < 6.  This is because of ES6 language features used.  Working around those to be ES3.1-compatible shouldn't be a problem if this looks largely good.

dougwilson:
Hi @mikesamuel makes sense. Yes, ideally if this can work on all Node.js versions, that would be best, because we strive to support as many Node.js versions as possible since this is a very low-level driver module (as compared to a higher level ORM, etc.).

I would, though, say that the actual formatting code should _not_ live in this module, otherwise it cannot be shared with the other MySQL modules like "mysql2" etc. The module https://github.com/mysqljs/sqlstring is dedicated to hosting the code for formatting.

mikesamuel:
Ok.  So to move this forward, I could move Template.js and its tests into a separate npm module.
Could that live under the `github.com/mysqljs` org?

On failing gracefully on older node runtimes, I could define a utility file:

```js
var calledAsTemplateTagQuick = (function () {
  try {
    return require('template-tag-common').calledAsTemplateTagQuick;
  } catch (ignored) {
    // Occurs if an ES6 parser is unavailable which is true for Node <= 6
    // String templates are not available there either, but might mismatch
    // if the calling code is transpiled, but mysql and its deps aren't.
    return function () { return false; };
  }
}());
```

If we don't attempt to load the interpolation handling code in branches that
don't pass that function, we should never observe that.

Test code would have to do an explicit version test when deciding whether to skip
tests and wrap template tag uses inside `new Function("...")`.

dougwilson:
Why not into the `sqlstring` module?

mikesamuel:
> Why not into the sqlstring module?

Sorry, didn't read closely enough.  Will do.

mikesamuel:
@dougwilson I filed https://github.com/mysqljs/sqlstring/pull/29 there.   If you want to close this, I can see what happens there and then come up with a new PR here, or if you want to leave this open, I can layer another commit on top once I'm ready to integrate a new version of sqlstring.

PR #1937, opened 2018-02-13
Added support for COM_INIT_DB in FakeServer.js

MayamaTakeshi:
Hi,
I use FakeServer.js to do automatic regression test generation for some applications that I was assigned to but that the previous devs didn't bother to write functional tests to.
One of such applications issued a COM_INIT_DB command that FakeServer.js doesn't support which was added in this commit.

### DIFF
--- a/lib/protocol/packets/ComInitDbPacket.js
+++ b/lib/protocol/packets/ComInitDbPacket.js
@@ -1,0 +1,15 @@
+module.exports = ComInitDbPacket;
+function ComInitDbPacket(database_name) {
+  this.command = 0x02;
+  this.database_name     = database_name;
+}
+
+ComInitDbPacket.prototype.write = function(writer) {
+  writer.writeUnsignedNumber(1, this.command);
+  writer.writeString(this.database_name);
+};
+
+ComInitDbPacket.prototype.parse = function(parser) {
+  this.command = parser.parseUnsignedNumber(1);
+  this.database_name     = parser.parsePacketTerminatedString();
+};
--- a/lib/protocol/packets/index.js
+++ b/lib/protocol/packets/index.js
@@ -20,3 +20,4 @@ exports.RowDataPacket = require('./RowDataPacket');
 exports.SSLRequestPacket = require('./SSLRequestPacket');
 exports.StatisticsPacket = require('./StatisticsPacket');
 exports.UseOldPasswordPacket = require('./UseOldPasswordPacket');
+exports.ComInitDbPacket = require('./ComInitDbPacket');
--- a/test/FakeServer.js
+++ b/test/FakeServer.js
@@ -302,6 +302,12 @@ FakeConnection.prototype._parsePacket = function(header) {
         this._handleQueryPacket(packet);
       }
       break;
+    case Packets.ComInitDbPacket:
+      if (!this.emit('init_db', packet)) {
+        this._sendPacket(new Packets.OkPacket());
+        this._parser.resetPacketNumber();
+      }
+      break;
     case Packets.ComPingPacket:
       if (!this.emit('ping', packet)) {
         this._sendPacket(new Packets.OkPacket());
@@ -370,6 +376,7 @@ FakeConnection.prototype._determinePacket = function(header) {
   var firstByte = this._parser.peak();
   switch (firstByte) {
     case 0x01: return Packets.ComQuitPacket;
+    case 0x02: return Packets.ComInitDbPacket;
     case 0x03: return Packets.ComQueryPacket;
     case 0x0e: return Packets.ComPingPacket;
     case 0x11: return Packets.ComChangeUserPacket;


lib/protocol/packets/index.js
```diff
@@ -20,3 +20,4 @@ exports.RowDataPacket = require('./RowDataPacket');
 exports.SSLRequestPacket = require('./SSLRequestPacket');
 exports.StatisticsPacket = require('./StatisticsPacket');
 exports.UseOldPasswordPacket = require('./UseOldPasswordPacket');
+exports.ComInitDbPacket = require('./ComInitDbPacket');
```
dougwilson: The contents of this file are in alphabetical order if you can reorder this line to match :+1:

lib/protocol/packets/ComInitDbPacket.js
```diff
@@ -0,0 +1,15 @@
+module.exports = ComInitDbPacket;
+function ComInitDbPacket(database_name) {
+  this.command = 0x02;
+  this.database_name     = database_name;
```
dougwilson: I think the equal signs on these two lines are accidentally unaligned.

lib/protocol/packets/ComInitDbPacket.js
```diff
@@ -0,0 +1,15 @@
...
+}
+
+ComInitDbPacket.prototype.write = function(writer) {
```
dougwilson: Please add a function name :+1:

lib/protocol/packets/ComInitDbPacket.js
```diff
@@ -0,0 +1,15 @@
...
+  writer.writeUnsignedNumber(1, this.command);
+  writer.writeString(this.database_name);
+};
+
+ComInitDbPacket.prototype.parse = function(parser) {
```
dougwilson: Also function name here too.

lib/protocol/packets/ComInitDbPacket.js
```diff
@@ -0,0 +1,15 @@
...
+  this.command = parser.parseUnsignedNumber(1);
+  this.database_name     = parser.parsePacketTerminatedString();
```
dougwilson: Align equals here too.

lib/protocol/packets/index.js
MayamaTakeshi: Hi,
No, the changes I did were just for FakeServer.js. It was not my intention to expose COM_INIT_DB in the client API as I don't think it is really necessary. 
The situation is this: I was put in charge of maintaining a python application but there were no regression tests written for it. So I came up with the idea of writing a tool to automatically write regression test for it by capturing all mysql interaction but for this I needed a library for the mysql protocol. So I checked this module and realized I could just use the FakeServer. But the python app does a (probably unnecessary) call to mysql_select_db (that issues COM_INIT_DB) and so my tool was failing.

I can add a method init_db in the client API just for the FakeServer to be testable but I think no client code will actually use this and so I would not like to pollute the interface with it. But if you think it is OK, I can do it.

dougwilson: How are you even using the FakeServer? We don't publish that file to npm and it is subject to sudden drastic changes because it's private to our use-case. The "mysql2" module actually provides a server implementation that is real and supported. Is that not what you should be using?

The only changes accepted to FakeServer.js are those that are directly related to the functionality in this module; as soon as something in FakeServer.js is not used in a test here, I delete it, and this PR would just end up immediately deleted since there are no tests making use of the functionality in FakeServer.js

dougwilson:
The module https://www.npmjs.com/package/mysql2 is what you want to use to create fake servers to call against in JavaScript. Here is an example from the repo: https://github.com/sidorares/node-mysql2/blob/master/examples/server.js

MayamaTakeshi:
Hi, the changes I did were only for FakeServer as a library itself. It was not my intention to expose COM_INIT_DB in the client API.
The situation is:
I was put in charge of a python application but no regression tests were ever written for it.
So before starting doing changes, I decided to write a tool to generate regression tests for it. For this I needed a mysql protocol library so I checked your module and realized I could use FakeServer for this.
But the python app does a (probably unnecessary) call to mysql_select_db (that issues COM_INIT_DB) and so my tool was failing.

I can add a method init_db to the client API so that FakeServer can be testable for it, but I think nobody would ever use this so I would not like to pollute the interface. But if you think this is OK, I can do it.

dougwilson:
The module https://www.npmjs.com/package/mysql2 is what you want to use to create fake servers to call against in JavaScript. Here is an example from the repo: https://github.com/sidorares/node-mysql2/blob/master/examples/server.js

lib/protocol/packets/index.js
MayamaTakeshi: Hi, I use yarn specifying the URL for the github repo in package.json:
  "mysql": "https://github.com/mysqljs/mysql#v2.15.0"
Yarn gets the test folder.
Sorry for the noise here and thanks for the tip about mysql2.

dougwilson: Very interesting. That looks like it just clones our repo. I guess that means from there, you're just requiring the file directly. Hmm, not sure if it's possible we can hide our one-off testing code from that sort of loading, though we don't document using that anywhere. When I first slapped that code together, the "mysql2" module didn't have server code, but now it does. Eventually that would be removed and "mysql2" or another module used and your code would probably end up broken at some point.

This module's code is also MIT licensed, so you're free to copy and paste it into your own repo to use at your liking as well.

MayamaTakeshi: Yes. Here is the code:
  https://github.com/MayamaTakeshi/mysql_query_collector/blob/master/src/index.js
(sorry, no documentation as this is ongoing work not for public consumption yet)
Thanks.

PR #1962, opened 2018-03-14
Fix MySQL 8.0.x incompatibilities

### LINKED ISSUE #1959: MySQL 8 incompatibilities
I work at the MySQL connectors team at Oracle and thought about running some tests using this module and the latest MySQL 8 series versions to check for incompatibilities.

There are a couple of things that need to change if there are plans to support those versions.

First of all, MySQL 8.0.4 introduces a new default authentication plugin - [`caching_sha2_pasword`](https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html). This is a breaking change and it will affect the entire handshake process, which currently does not work for new user accounts created using this plugin or, most importantly, not using a specific plugin at all (since it's now the default option). User accounts created with the [`mysql_native_password`](https://dev.mysql.com/doc/refman/8.0/en/native-pluggable-authentication.html) plugin will still work though.

Additionally, MySQL 8.0.2 disables the `local_infile` system variable by default, which means that `LOAD DATA LOCAL INFILE` statements require the user to enable that feature explicitly (using `SET GLOBAL local_infile = true`). I guess this is mostly a documentation concern, however there are a couple of tests
that are failing, so I thought it was also worth mentioning.

I'm working on a pull request to address these issues, but before moving on, I wanted to understand if the goal is to eventually support the MySQL 8 series or not, and what's the policy regarding non-GA versions, which is still the case for MySQL 8.0.4 (although it's already in RC, which might mean something). This last point is important because there are breaking changes between DMR releases and I wanted to understand if the patch(es) would need to factor those in.

ruiquelhas:
This patch accommodates some breaking changes introduced with MySQL 8.

Closes #1959 

In a nutshell, the `caching_sha2_password` plugin (which is used by default since MySQL 8.0.4) hashes the password using SHA-256 and, after a first successful authentication attempt, saves it in a cache. That first attempt needs to be done under one of two conditions. The client either uses SSL and sends the password as clear text or it encrypts the password using the RSA public key generated by the server. On any subsequent attempts, until the password is somehow removed from the cache or the server shuts down, these rules no longer apply.

The handshake process remains unchanged when connecting to any server with version lower or equal to `8.0.3`. Whereas for `8.0.4` or above, the process is now the following:

- the client sends a `ClientAuthenticationPacket` with a scramble computed using a SHA-256 hash
- if the password is not cached, the server sends back a `PerformFullAuthenticationPacket`
- if the client uses SSL, the password is sent to the server (as clear text) via a `ClearTextPasswordPacket` to which the server replies with a `OkPacket`
- otherwise it uses the server authentication public key compute the scramble, sending a `AuthSwitchResponsePacket` to which the server replies with a `OkPacket`
- if the client does not know the server public key (is not provided by the user), it requests it from the server, which sends it back using a `AuthMoreDataPacket`
- after a first successful authentication attempt, and until the password is cached, the server will reply to the initial `ClientAuthenticationPacket` with a `FastAuthSuccessPacket` (which basically just signals that an `OkPacket` will follow)

If the account is created using the `mysql_native_password` authentication plugin, the client will just fall back to the "traditional" process during the handshake, keeping compatibility, by default for any previously supported server version.

MySQL 8.0.2 disables the `local_infile` server variable by default, which breaks a couple of integration tests. The tests were updated to enable the feature by themselves (something that does not have any effect on older server versions and allows the tests to pass with newer versions).

Additionally, one of the integration tests was updated to avoid failing after the first run (using any server version) since it tried to create a table that already existed from the previous runs.


dougwilson:
This is awesome, thank you soo much! The only real issues I'm seeing with the PR to continue to work on is the following:

- [ ] Get support working on Node.js 0.10 and below, or we can decide if this PR should be shelved for the next major release as breaking change to drop those Node.js versions. Being a widely-used base driver we try to support as many Node.js and MySQL versions as possible at once to make it easy to use for everyone.
- [ ] There may not be enough coverage of the new code added, as the coverage with the PR is decreasing. There are probably some areas that can use tests.
- [ ] MySQL 8 image needs to be stood up on the CI server to test against. This is done in the `.travis.yml` file similar to our current MySQL / MariaDB matrix.

lib/ConnectionConfig.js
```diff
@@ -106,14 +108,19 @@ ConnectionConfig.getDefaultFlags = function getDefaultFlags(options) {
...
-    '-PLUGIN_AUTH',       // Does *NOT* support auth plugins
+    '+PLUGIN_AUTH',       // Supports auth plugins
     '+PROTOCOL_41',       // Uses the 4.1 protocol
     '+PS_MULTI_RESULTS',  // Can handle multiple resultsets for COM_STMT_EXECUTE
     '+RESERVED',          // Unused
     '+SECURE_CONNECTION', // Supports Authentication::Native41
     '+TRANSACTIONS'       // Expects status flags
   ];
 
+  if (options && options.database) {
```
dougwilson: Is this change related specifically to the MySQL 8 support, or just a general fix. Just curious.

ruiquelhas: For the new authentication magic to happen in the server, we need to explicitly provide the authentication plugin name in the `ClientAuthenticationPacket`, and, doing that, we need to be careful of which flags to enable, otherwise the server will, for instance, interpret `caching_sha2_password` or `mysql_native_password` as the schema/database name, if none is provided.

I didn't want to mess around with the `PacketWriter` and this seemed the easiest way to overcome that hurdle. However, we can try to think of something else if it feels weird.

dougwilson: Sure, but I was more curious on if this is a general fix or specific to MySQL 8. But even then, users can pass in their own `flags` list, and this module will enable / disable them unless it's listed here with a flag to force one way or the other. If having it set when it shouldn't is an issue, then it should probably be flagged here so the users cannot pass in the flag and get it set when no db is specified.

ruiquelhas:
To address your points.

Regarding the first one, I've missed those tests (sorry), but as far as I can tell there is no sane way to use RSA public key encryption on `node <=v0.12.0`. There are some packages on npm that try to add support through native addons, but I think that might not be a good idea here, since it's pure JavaScript. So the only easy way to work around that is to require users on those older `node` versions to use SSL when connecting to new MySQL versions. If you are OK with that, we can go that route, otherwise, going in the next major version is probably the best possibility (I guess that's mostly your call).

Missed the coverage as well, I'll try to fix that ASAP.

Again, missed the CI config. Should be fine, at least on Travis. However, besides that, I suggest we also create a test account to run the tests (besides `root`) to make sure the handshake works fine for both empty and non-empty passwords, which lead to different behaviors. Any issue with that? Other than that, I guess on windows we are out of luck anyway since appveyor only seems to support MySQL 5.7, correct?

dougwilson:
> Regarding the first one, I've missed those tests (sorry), but as far as I can tell there is no sane way to use RSA public key encryption on node <=v0.12.0. There are some packages on npm that try to add support through native addons, but I think that might not be a good idea here, since it's pure JavaScript. So the only easy way to work around that is to require users on those older node versions to use SSL when connecting to new MySQL versions. If you are OK with that, we can go that route, otherwise, going in the next major version is probably the best possibility (I guess that's mostly your call).

Ah, very sad :( Did you do a through search on this (it sounds like it)? Because I can always help if you want to focus on the other parts, but just asking before I duplicate the search for no reason. If we really have to drop supported versions, we can, and it wouldn't particularly delay this, just wanted to make sure it was actually required to do so before saying we should :)

> Again, missed the CI config. Should be fine, at least on Travis. However, besides that, I suggest we also create a test account to run the tests (besides root) to make sure the handshake works fine for both empty and non-empty passwords, which lead to different behaviors. Any issue with that? Other than that, I guess on windows we are out of luck anyway since appveyor only seems to support MySQL 5.7, correct?

Yea, so in Travis CI it just uses a Docker image. You can make multiple non-root accounts too; the current setup was just what was needed for testing, so if MySQL 8 requires more elaborate setup, that sounds fine to me. I would ignore AppVeyor.

ruiquelhas:
I'm using `crypto.publicEncrypt()` which was added in `v0.11.14`, and it uses native bindings which I assume were also added around that time. There are some libraries on npm (e.g. [ursa](https://github.com/JoshKaufman/ursa) and [dcrypt](https://github.com/dekz/dcrypt)) that claim support for earlier node versions, however they seem to implement their own native bindings, which I guess would, sort of, defeat the purpose of this module (or maybe not, you tell me). 

There's this [node-rsa](https://github.com/rzcoder/node-rsa) thingy which apparently is pure JS, but I couldn't make it work and it's pretty slow. I've also seen other [approaches](https://github.com/ragnar-johannsson/rsautl) like delegating work to openssl via a child process, but that's probably not a good idea either, since it introduces a huge moving part. There's also this [forge](https://github.com/digitalbazaar/forge) kit which brings the entire browser kitchen sink with it and I also haven't managed to figure out how it worked.

So, no good news from my side. However, if you can do your own research, that would, of course, be great. However, I'm not a cryptography expert, but from what I can tell, the reason I have not found anything good is because this RSA thingy in pure JavaScript is still not really feasible (`number` precision, primes and what not).

My suggestion here is to fail with a client error (the `HANDSHAKE_SECURE_TRANSPORT_REQUIRED` introduced by this PR) when `process.env.version` tells us that we are below `v0.11.14`, which basically would be the default behavior using `{ ssl: false, secureAuth: false }`. It's a pretty naive solution but at least it's not really a breaking change, because existing implementations will keep working as expected with MySQL 5.7 and, let's be honest, people wo don't upgrade from those ancient node versions will, almost for sure, not upgrade MySQL to series 8 as well.

However, if you still feel you are better off bumping the major version, that's fine.

ruiquelhas:
@dougwilson what's your preference about the issue I mentioned? Do you want me to add the workaround for older node.js versions or do you think it's better to have a major version bump (in that case I guess all we need is to remove the CI setup for older versions)?

dougwilson:
Sorry I didn't reply. Your comment on a new issue reminded me of this. Yea that works: if the server requires an auth that the client can't do we can error out. We should feature detect that, though, not sniff the version string.

ruiquelhas:
Ok. Just to be sure. Are you talking about sniffing the Node.js version string, MySQL version string, or both? Because we have two different issues right now. 

Supporting MySQL 8.0.4 (previous DMR release) which expects the password to be encrypted with the server public key using `RSA_PKCS1_PADDING`, whereas since 8.0.11 (the new GA release), the encryption should use `RSA_PKCS1_OAEP_PADDING`. We can always retry the handshake, but the only way to avoid that is by "sniffing" the MySQL server version.

Regarding the Node.js encryption APIs, I guess we can always resort to duck typing to check if they are available (or not) instead of sniffing the version. So, no problem here.

Let me know if there is something else that bothers you about these or other issues. I'll try to push this forward ASAP. Maybe during the weekend.

dougwilson:
Why do you have to sniff the server version? Shouldn't the server say which one it expects directly? I'm just asking because I think version numbers break down when we start talking about compatible things like MariaDB, right? They may share some version numbers.

I have done Node.js version sniffing before and when it broke the Node.js collaborators just said I should never have been version sniffing their versions and told me feature detect only. So if that is the official direction from there I would go with that.

johannes:
The case where we might need to sniff the server version is to distinguish MySQL 8.0.10 (pre GA, first one introduced chaching_sha1_passwors) from MySQL 8.0.11 (GA). since 8.0.11 was released yesterday one also could not support 8.0.10, as nobody should be using that RC anyways.

jojannes

dougwilson:
I like the idea to either (a) just ignore the pre GA padding or (b) retry with the old padding because that would not involve the version sniffing :+1:

ruiquelhas:
@dougwilson I've kept the option of providing a custom key padding, to allow users to exceptionally connect to a MySQL `8.0.4 RC` server. In any case, the default value is always `RSA_PKCS1_OAEP_PADDING` which means it will work, by omission, with the latest `8.0.11 GA` version.

I couldn't find a sane way to add an additional Travis CI test matrix for users with non-empty passwords though. I was constantly bumping into an authentication error (I believe it might be related to the `mysql` docker setup and environment variables mixup), but maybe you can give it a try.

Let me know if there is something else you would like to see in this PR. Otherwise, feel free to merge it! :wink:

dougwilson:
This is awesome, thank you. I've been reviewing it, and sorry I didn't reply that I was in progress. I did find at least one regression just in existing behavior so far, and pushed a test to cover this to `master` so we have it covered for the future as well. I noticed the change for the `ChangeUser` sequence no longer passes any of the options down to the `Sequence` constructor, which differs not from every other sequence. Not sure what the purpose of that change was, but it at least prevents the `timeout` option from getting passed down.

I'll look at what the fix is and push it up to your PR branch here 👍

fan123199:
The merge process last so long, and the pr is still not merged. So it there any alternative way to solve the problem of "ER_NOT_SUPPORTED_AUTH_MODE: Client does not support authentication 
authentication "? I am nodejs startup and just want to connect to mysql 8.0 for demo.

ruiquelhas:
@fan123199 everything should still work if you use the `mysql_native_password` plugin.

```sql
ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'YourRootPassword';
-- or
CREATE USER 'foo'@'%' IDENTIFIED WITH mysql_native_password BY 'bar';
```

Slyke:
In addition to what @ruiquelhas said, I also had to:
```
mysql> FLUSH PRIVILEGES;
Query OK, 0 rows affected (0.00 sec)
```
And then restart the server.
If you are using multiple MySQL users, ensure that you update all of them. In my case I also had to use `'%'` instead of `'localhost'`.

affanshahid:
Any updates on this? Unable to authenticate with MySQL 8

ruiquelhas:
@Osuriel

> Is this a good fix that can be deploy to production or is this just a workaround for getting things working locally?

The new default authentication plugin is more secure because it's using SHA256 instead of SHA1 to hash the password, but it's not available for MySQL versions below 8.0, which means most existing MySQL production instances are not using it as well. So, it all boils down to how much you can trust SHA1 (for which collision attacks have already been performed in 2017). I guess having a good password should help to mitigate or even prevent that.

> I am also assuming that what is happened is that 8.0.11 MySQL Community Server is just not compatible with the npm mysql library right out the box at the moment. To make it compatible right out the box You have submitted this bug fix but it hasn't been merged to master yet, and that is the reason it still doesn't work for me right out the box even tho i have the latest version of the mysql dependency.
>
> Is my assumption correct?

Correct.

> If I am correct about my assumption, is there a practical way to implement the bug fix to my codebase before it is merged into master? and if there is, is that something that would be advised, or good practice?

I don't know if it is a good practice or not, it mostly depends on your internal policies and processes (and how much you would be "trusting" the fork in this case), but you can install and link npm packages from a git repo even using a specific branch or commit.

```bash
$ npm install ruiquelhas/mysql#feature/caching-sha2-password
```

FrancescoDiMaggio:
Any update? Still unable to connecto to MySQL 8

Osuriel:
@ruiquelhas thank you so much! You just helped me become a better developer hannibwas yesterday! 💪🏽🔥

kibertoad:
@dougwilson Is this still planned to be merged? It obviously keeps accumulating conflicts while it is not :(

dougwilson:
This PR regresses behavior and cannot be merged until fixed as I noted above. I volunteered to fix the PR but have not gotten to it. Anyone is welcome to fix the regression in the PR, though, and will merge.

kibertoad:
@dougwilson Thanks for the update! Not sure I'll get to it in the nearby future personally, though (but it's a possibility).

dougwilson:
It's np. I did get it mostly rebased and fixed locally but didn't quite finish. I'll bring this close to the top of my current todo list.

ruiquelhas:
> I'll look at what the fix is and push it up to your PR branch here

@dougwilson I was under the impression you would take over from there, and just assumed something else was blocking this. In any case, I'll gladly help to push this forward. Please, let me know what I can do.

ruiquelhas:
Alright, perfect. Let me know if there is something I can help with.

ktjd123:
I think you should keep track of this issue :) still on 8.0.

iedmrc:
Hi all,
Thanks for your effort!
But please merge this PR. It's been around 8 months that we have this compatibility issue with mysql 8.
Thanks 🎉

milespratt:
Thanks to all who have worked hard on the `caching_sha2_password` support. Are there plans to merge this soon?

iedmrc:
> Hi all,
> Thanks for your effort!
> But please merge this PR. It's been around 8 months that we have this compatibility issue with mysql 8.
> Thanks 🎉

I  don't get what is the wrong with this comment. It was pretty gentle. But It seems, you spend all the time for marking comments as something, not try to solve this BIG ISSUE which isn't solved for almost 10 months!

Issue #1978, opened 2018-04-02
Explain the Forms about Single Placeholder on Readme.md


anliting:
This is good: `cn.query('select ?',[0],()=>{})`

It seems that this also works: `cn.query('select ?',0,()=>{})`

I found that there is an example using the later form:
```
var post  = {id: 1, title: 'Hello MySQL'};
var query = connection.query('INSERT INTO posts SET ?', post, function (error, results, fields) {
  if (error) throw error;
  // Neat!
});
console.log(query.sql); // INSERT INTO posts SET `id` = 1, `title` = 'Hello MySQL'
```

Is this thing well-documented in Readme.md?

dougwilson:
I think so, but if you feel there are any gaps, please feel free to submit a pull request with the changes you think should be made :+1:

anliting:
I don't see that `cn.query('select ?',[0],()=>{})` can be written as `cn.query('select ?',0,()=>{})` explained in English in `Readme.md`. If there is such an explanation, pointing it out for me would help.

If there is really no such an explanation, I think I'll make a PR in a few days.

dougwilson:
I'll take a look when I get to a computer soon :+1: but if you don't see it, I'm sure you already read through the readme and didn't find it, do feel free to make a PR :+1:

anliting:
I need to check another thing about this to make a PR: The second form can only be used when the only argument is not an array, right? i.e. `mysql.format('select ?',[[0,1]])` cannot be written as `mysql.format('select ?',[0,1])`.

dougwilson:
I'll double-check when I get to a computer 👍 not 100% sure off-hand.

anliting:
I found that it also doesn't work when `values` is `[undefined]` or `[null]`.

dougwilson:
Documentation for this was added in PR #1979 . I forgot to close this issue when merged, but linking them together to close it out. Thanks again for the pull request 👍

PR #1979, opened 2018-04-02
Explain the Forms about Single Placeholder on Readme.md (#1978)

anliting:


anliting:
@dougwilson Does it look good?

Issue #1990, opened 2018-04-07
Row Column Mappings are Wrong


Crowes:
I've been facing an issue where one of my `INT` columns was returning as `NaN`, even just as a raw `RowDataPacket`. While I do some custom type casting (for Dates & Boolean), the majority of it is done by the baked-in parser.

Debugging the parser in `RowDataPacket.js` shows that the value it returns for my column is actually that of a different column entirely.

Here, in the debugger (WebStorm), we can see that the `field.name` is `userPrivilege`, and the value is `2017-12-16 14:19:23`. 
https://i.imgur.com/OyWOxjw.png

Here in the table (DataGrip), we can see that the value of `userPrivilege` for this record is `3`, not the date. That date is actually inside `userCreation`.

https://i.imgur.com/eExOstf.png

dougwilson:
Very strange indeed! Definitely I want to take a look and fix this for you. I don't have enough information in order to figure out what is wrong yet. Can you provide as much of the following as possible?

1. Version of this module you are using.
2. Version of Node.js you are using.
3. Distribution name and version of the MySQL server you are using.
4. DDL of the table (or tables of this involves a join) that is being queried.
5. The query that is being run on the table(s).
6. Example data to populate into the table(s) that causes the query to return mixed up columns.

Thank you!

Crowes:
I'll do my best to clarify a bit. 

I've been using an [url=https://github.com/Gallink/Joints/blob/master/parts/DatabaseCasting.js]abstraction[/url] for my type casting, which no longer works. When I reverted to my [url=https://i.imgur.com/aXPjwkH.png]old casting logic[/url] (changed in [url=https://github.com/Gallink/Joints/commit/f1286d3772d58ad837053b37fbc374f349887152]this commit[/url]). I can't spot why -- there's no difference besides placement.

**Version of this Module**: `2.15.0`
**Version of Node.js": `8.1.2` (may be due an update)
**Distribution & Version of MySQL**:  `5.7 Community`
**DDL of the Table**:  https://i.imgur.com/P6ggQFu.png
**Query Being Ran**:  `SELECT * FROM user WHERE userId = ? LIMIT 1` with `["1"]`
**Example Data**: 

      INSERT INTO user (userEmail, userName, userForename, userSurname, userPassword, userDate, userCreation, userToken, userPrivilege, userEnabled)
[...]

PR #1991, opened 2018-04-07
Clarified the type cast parser call warning.

Crowes:
Answering that "Why can i not call it more than once?" question. Refer to this issue: https://github.com/mysqljs/mysql/issues/1990


dougwilson:
I never got a response back and the fork repo is deleted, so it won't let me merge. You're welcome to open a new PR at any time 👍

PR #2029, opened 2018-05-29
updated pool.release() to pool.releaseConnection()

abhimassive:
updated pool.release() to pool.releaseConnection()

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -351,7 +351,7 @@ pool.getConnection(function(err, connection) {
 });
 ```
 
-When you are done with a connection, just call `connection.release()` and the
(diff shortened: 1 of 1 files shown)


sidorares:
are you sure it's correct change @abhimassive ?

.releaseConnection is `Pool`s method, `connection.release()` calls it

dougwilson:
Yep, there is no releaseConnection method on a connection object, just like there is no release method on a pool object. OP may have had them mixed up in his or her code. OP if you want to create a new issue describing your issue along with full example code demonstrating it, we can help :+1:

PR #2058, opened 2018-07-02
Improved readability in Readme.md

ealain:
I removed a few spurious words from a paragraph in Readme.md. I assume these were inadvertently left in a previous change.


ealain:
Do you think the below wording would be acceptable?

>The `end` method takes an _optional_ callback that you can use to know when
all the connections are ended.
**Once `pool.end` is called, `pool.getConnection` and other operations
can no longer be performed.** Wait until all connections in the pool are
released before calling `pool.end`. If using the shortcut method
`pool.query`, wait until it completes.
`pool.end` calls `connection.end` on every active connection in the pool.
This queues a `QUIT` packet on the connection and sets a flag to prevent
`pool.getConnection` from creating new connections. All commands / queries
already in progress will complete, but new commands won't execute.

Readme.md
```diff
@@ -455,28 +455,18 @@ pool.end(function (err) {
...
-prevent `pool.getConnection` from continuing to create any new connections.
-
-Since this queues a `QUIT` packet on each connection, all commands / queries
-already in progress will complete, just like calling `connection.end()`. If
-`pool.end` is called and there are connections that have not yet been released,
-those connections will fail to execute any new commands after the `pool.end`
-since they have a pending `QUIT` packet in their queue; wait until releasing
-all connections back to the pool before calling `pool.end()`.
-
-Since the `pool.query` method is a short-hand for the `pool.getConnection` ->
```
dougwilson: Can you find a spot to keep the note about it being a short-hand for the three methods? A user previously added that part to this section because they were confused without it. I would like to keep it there for them if possible.

ealain:
Indeed, it makes sense to add this reminder here.

PR #2062, opened 2018-07-06
There is an issue inside catch block of parser

divsbhalala:
I'm using "loopback-connector-mysql" and this connector use this library, So here I request you to review this fix or check if there is some other better way to handle this error, please.
I'm facing this throw exception error much time and this will always crash the node.js server process


dougwilson:
Please make sure to update the PR to follow our contribution guidelines https://github.com/mysqljs/mysql/blob/master/Readme.md#contributing

dougwilson:
Closing due to no response.

PR #2102, opened 2018-09-18
Remove travis ci support for legacy node versions

amilajack:


dougwilson:
As a base driver we support these versions.

PR #2124, opened 2018-10-20
Query Observables

HassanAzzam:
This pull request adds new feature 'Query Observables'.
It uses ReactiveX Observables as an interface for retrieving data instead of callbacks way.

Observables has replaced Promises and getting more used, I think a feature like this will be satisfying for many developers.

The pull request has test cases for the new added methods and they run as expected.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -4,6 +4,12 @@ This file is a manually maintained list of changes for each release. Feel free
 to add your changes here when sending pull requests. Also send corrections if
 you spot any mistakes.
 
+## v2.16.1 (2018-10-20)
+
+* Add `rxjs ^6.3.3` to dependencies
+* Add `connection.queryObservable()`
+* Add `pool.queryObservable()`
+
 ## v2.16.0 (2018-07-17)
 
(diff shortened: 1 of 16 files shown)


dougwilson:
Nice! I will take a look though it soon. A couple comments currently: (1) please take a look into the CI failures and (2) please remove the rxjs dependency, as they declare a requirement of npm >= 2.0.0 but we support npm 1.x. 

Making such a platform requirement change would be a semver major for this module and until just the other day I never heard of it, so it doesn't seem like there is any demand besides you to justify dropping support for a platform. Is there a way to either (a) add this support without dropping npm 1.x support or (b) can tis be implemented as an add on module instead of directly in this core driver module?

HassanAzzam:
I downgraded 'rxjs' to an older version so it supports Node v0.6 and NPM v1.1 and all tests run successfully locally.

Check case 11 in travis-ci, it's not something related to my edits.

HassanAzzam:
I can't remove this dependency, everything is about it at the first place.

I will close the pull request for now until there is a proper way to support all versions.

PR #2129, opened 2018-11-04
Add hints for yarn install

gtzilla:
The docs and readme are very thorough. Very nice! I spent a little extra time verifying that the yarn package with name `mysql` was the same and that the `mysqljs/mysql` install worked and that `--flat` would work as expected.  

commit message: Install with yarn. Confirm that the commands point to the correct package. And acknowledge additional audience/user support.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -69,10 +69,10 @@ Installation is done using the
 $ npm install mysql
 ```
 
-You can also install mysqljs using the [Yarn package manager](https://yarnpkg.com/lang/en/docs/install/). Yarn is an alternative to npm, which enables a flat dependency tree.
+You can also install this module using the [Yarn package manager](https://yarnpkg.com/lang/en/docs/install/). Yarn is an alternative to npm, which enables a flat dependency tree.
 
 ```sh
-$ yarn add mysql --flat
+$ yarn add mysql
 ```
 
 For information about the previous 0.9.x releases, visit the [v0.9 branch][].
@@ -84,10 +84,10 @@ if a bugfix is working. In this case, please do:
 $ npm install mysqljs/mysql
 ```
 
-For Yarn, it look as follows:
+For Yarn, it would be the following command:
 
 ```sh
-$ yarn add mysqljs/mysql --flat
+$ yarn add mysqljs/mysql
 ```
 
 
--- a/Readme.md
+++ b/Readme.md
@@ -69,6 +69,12 @@ Installation is done using the
 $ npm install mysql
 ```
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Awesome, thank you! Your comments make it sound like `--flat` could have potentially not worked. Is there something we need to do to ensure it continues to work so we know if we have to update the README? If so, is it possible the CI can check this so we'll know right away instead of hoping a `yarn` user will report it (and perhaps confuse new `yarn` users in the process why the install command doesn't work)?

gtzilla:
I find some projects end up having many resolutions / conflicts due to the dependency tree. I wasn't asked to resolve anything when I tested, which is great. It's not needed generally. I just assume people not using npm are doing so for similar reasons regarding the dependency tree. If it's a point of confusion it should be removed.

Readme.md
```diff
@@ -69,6 +69,12 @@ Installation is done using the
 $ npm install mysql
 ```
 
+You can also install mysqljs using the [Yarn package manager](https://yarnpkg.com/lang/en/docs/install/). Yarn is an alternative to npm, which enables a flat dependency tree.
```
dougwilson: `mysqljs` is the name of the GitHub organization, not the module/product. Just call it `mysql` or "this module" here.

Readme.md
```diff
@@ -69,6 +69,12 @@ Installation is done using the
...
+
+```sh
+$ yarn add mysql --flat
```
dougwilson: See comment about `--flat` in comments.

Readme.md
```diff
@@ -78,6 +84,13 @@ if a bugfix is working. In this case, please do:
 $ npm install mysqljs/mysql
 ```
 
+For Yarn, it look as follows:
```
dougwilson: I would change "it look as follows" to "it would be the following command".

gtzilla:
@dougwilson thanks for the fixes.

gtzilla:
I'm slightly unfamiliar with the process to resolve and close this ticket. What steps do I need to take?

dougwilson:
What about adding yarn to CI so we can ensure the module stays compatible? Not sure where that suggestion landed.

dougwilson:
Ok, I'm going to close this for now, but you can always re-open it. I'm very hesitant to add `yarn` to the readme, as I never use `yarn`, and this doesn't add anything to the CI to guarantee `yarn` compatibility. If this module just happens to work with `yarn` and `yarn` users know how to install things with it, then I don't think there is a specific reason to add this, especially by having it listed, it makes it sound like a supported platform, but there isn't any collaborators who have stepped forward to help ensure that support continues to function into the future (either manually or with a CI test).

PR #2195, opened 2019-03-14
Add connectionLifetime to pool options

### LINKED ISSUE #1276: Connections in pool have an infinite lifetime
I have a process running all the time and when it idles for a while the connections in the pool are in a sleeping state. Eventually MySQL purges those connections based on the `wait_timeout` setting in `my.cnf`. Once this happens and I try to use a connection it will fail because the module assumes the connection is still live and tries to use it only to get a timeout or connection exception.

I would propose adding an option when creating a pool `connection_lifetime` which will work in a following way. When getting a connection from the pool, if the `connection_lifetime` is not exceeded we simply return the connection, if however it is exceeded we kill the selected free connection and create a new one in it's place. That should solve the issue.

dicearr:
This option is specially useful when connecting to serverless databases, in which the instances are turned off when there is no connection alive.

Fixes #1276

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -12,6 +12,7 @@ you spot any mistakes.
 * Remove special case for handshake in determine packet code
 * Small performance improvement starting command sequence
 * Support Node.js 11.x
+* Add `connectionLifetime` to pool options
 
 ## v2.16.0 (2018-07-17)
 
--- a/Readme.md
+++ b/Readme.md
@@ -391,6 +391,9 @@ constructor. In addition to those options pools accept a few extras:
 * `queueLimit`: The maximum number of connection requests the pool will queue
   before returning an error from `getConnection`. If set to `0`, there is no
   limit to the number of queued connection requests. (Default: `0`)
+* `connectionLifetime`: The maximum number of milliseconds a connection can be
+  alive in the pool. If set to `0` connection will live until it is manually
+  destroyed or the pool is closed. (Default: `0`)
 
 ## Pool events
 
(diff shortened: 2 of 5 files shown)


dicearr:
Two tests are failing in one of the Travis jobs.

Node version: 6.16
DOCKER_MYSQL_TYPE=mariadb
DOCKER_MYSQL_VERSION=10.2

Files are:

`test/unit/pool-cluster/test-connection-error-remove.js`
`test/unit/pool-cluster/test-connection-rr-error.js`

I think those failures are unrelated to the changes but I am not hundred percent sure. The AppVeyor build for node v6.16 is passing.

dougwilson:
Thank you. I will check out the CI failures 👍

dolsup:
I think the timeout should be set every time after pool connection is released, not at the creation. and it have to be clear when pool connection is acquired again after release.

dicearr:
That would be something like `idleConnectionTimeout` rather than `connectionLifetime`, right? 

I have had time to check the PR again and you're completely right. In the way it is currently built if the connection expires while is being used the socket gets destroyed causing an error to be thrown. I think that unless I could gracefully close the connection, your suggestion is the way to go. Was that the reasoning behind your comment?

I have found #1749 which seems to be an attempt for implementing the `idleConnectionTimeout`, and #962 which is the related issue. The PR is stalled for no apparent reason.

@dougwilson Could you please share your thoughts about this? If the `idleConnectionTimeout` is the way to proceed I will close this PR and open a new one.

dolsup:
@dicearr You said exactly what I meant. 
In my case I implemented something like `idleConnectionTimeout` for my production which was built using microservice architecture and needs flexible pool size for each service. because the max connection capacity of my database is so small. I hope it goes to official feature.

And you can consult this PR.
https://github.com/sidorares/node-mysql2/pull/736

dicearr:
Closed in favor of #2218.

PR #2218, opened 2019-05-10
Add idleConnectionTimeout to pool options

### LINKED ISSUE #962: It would be good to add some idleTimeout to pool options

### LINKED ISSUE #1276: Connections in pool have an infinite lifetime

dicearr:
This option is specially useful when connecting to serverless databases, in which the instances are turned off when there is no connection alive.

Coming from #2195.

This aims to close #1276 and resolve #962.

### DIFF
--- a/lib/PoolConnection.js
+++ b/lib/PoolConnection.js
@@ -5,6 +5,8 @@ var Events     = require('events');
 module.exports = PoolConnection;
 inherits(PoolConnection, Connection);
 
+function _noop() {}
+
 function PoolConnection(pool, options) {
   Connection.call(this, options);
   this._pool  = pool;
@@ -34,7 +36,7 @@ PoolConnection.prototype.release = function release() {
 
   if (this._pool.config.idleConnectionTimeout) {
     this._idleTimeout = setTimeout(
-      this.destroy.bind(this),
+      this._realEnd.bind(this, _noop),
       this._pool.config.idleConnectionTimeout
     );
   }
--- a/lib/PoolConnection.js
+++ b/lib/PoolConnection.js
@@ -32,6 +32,13 @@ PoolConnection.prototype.release = function release() {
     return undefined;
   }
 
+  if (this._pool.config.idleConnectionTimeout) {
+    this._idleTimeout = setTimeout(
+      this.destroy.bind(this),
+      this._pool.config.idleConnectionTimeout
+    );
+  }
+
   return pool.releaseConnection(this);
 };
 
@@ -56,6 +63,11 @@ PoolConnection.prototype.destroy = function () {
 PoolConnection.prototype._removeFromPool = function _removeFromPool() {
   if (!this._pool || this._pool._closed) {
     return;
+  }
+
+  if (this._idleTimeout) {
+    clearTimeout(this._idleTimeout);
+    this._idleTimeout = null;
   }
 
   var pool = this._pool;
--- a/Changes.md
+++ b/Changes.md
@@ -7,6 +7,7 @@ you spot any mistakes.
 ## HEAD
 
 * Support Node.js 12.x #2211
+* Add `idleConnectionTimeout` to pool options
 
 ## v2.17.1 (2019-04-18)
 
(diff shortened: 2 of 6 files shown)


gyj1278:
This option is great，and when will it be released to the official edition?@@dougwilson

dougwilson:
I am working through this module prs and issues this week.

lib/PoolConnection.js
```diff
@@ -32,6 +32,13 @@ PoolConnection.prototype.release = function release() {
     return undefined;
   }
 
+  if (this._pool.config.idleConnectionTimeout) {
+    this._idleTimeout = setTimeout(
+      this.destroy.bind(this),
```
dougwilson: This should probably end the connection gracefully, running the QUIT packet to the server instead of just dropping the TCP connection. Various database setups will block a client that keeps dropping TCP connections without properly closing them.

akleiber:
@dicearr this would be ❤️ by many cloud users. Any plan to continue this PR?

dicearr:
Hi @akleiber. After the first review I tried to gracefully close the connection by using `conn.end` instead of `conn.destroy` in 6633202, which is a fixup commit. So, I don't think there is too much I can do here. If 6633202 is not what @dougwilson meant, some feedback would be appreciated. If it is, then I just need to squash the commits.

rolandmarg:
We use this library with aws aurora and idleConnectionTimeout is really needed. I don't really have time to wait for this ticket to be merged. What do you guys think about attaching a timer on connection 'release' event, which will close connection if it's not acquired after 'x' time, can be done on top of library.

Mwni:
This is a temporary solution until this functionality has been merged. Place this right after importing the mysql library.
```
(() => {
	let createPool = mysql.createPool

	mysql.createPool = function(config){
		let pool = createPool(config)

		if(config.idleConnectionTimeout){
			let getConnection = pool.getConnection

			pool.getConnection = function(callback){
				getConnection.call(pool, (err, connection) => {
					if(err){
						callback(err, connection)
						return
					}

					if(connection.__lastUsed){
						if(Date.now() - connection.__lastUsed > config.idleConnectionTimeout){
							console.log('nuke')
							connection.destroy()
							pool.getConnection(callback)
							return
						}
					}

					connection.__lastUsed = Date.now()
					callback(err, connection)
				})
			}
		}

		return pool
	}
})()
```
This is forward compatible. So what it does is patch the `createPool` function to check for `idleConnectionTimeout`, if this setting is present, it patches the pool's `getConnection` function to destroy any connection, that it was about to serve, that has been been idle for longer than specified.

olsavmic:
We have the very same use case with AWS Aurora (dynamic number of read replicas so in order to use both connection pool and discover new hosts, we need to close the connection after some time and recreate new with refreshed DNS).
It's basically required for HA.

Is there anything I can do it push this MR further in order to merge it? @dougwilson 
Thanks!

yqsailor:
https://github.com/mysqljs/mysql/pull/2507/files

yqsailor:
https://github.com/mysqljs/mysql/pull/2507/files

Hello, everyone, I commit a pr and to solve this issue, any feedback would be appreciated.

jpike88:
@dougwilson last commit on this repo was a year ago. Any plans to merge this?

PR #2224, opened 2019-05-28
Update Readme.md to clarify disconnect behavior.

datvong-wm:
Clarify the disconnect behavior with Pool, and the error POOL_NOEXIST.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -650,6 +650,13 @@ terminated, an existing connection object cannot be re-connected by design.
 With Pool, disconnected connections will be removed from the pool freeing up
 space for a new connection to be created on the next getConnection call.
 
+With PoolCluster, disconnected connections will count as errors against the
+related node, incrementing the error code for that node. Once there are more than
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -579,7 +579,7 @@ Re-connecting a connection is done by establishing a new connection. Once
 terminated, an existing connection object cannot be re-connected by design.
 
 With Pool, disconnected connections will be removed from the pool freeing up
-space for a new connection to be created on the next getConnection call.
+space for a new connection to be created on the next getConnection call.  If all connections die, the cluster node will be removed.  If a database is restarted or temporarily offline, a `POOL_NOEXIST` error may occur.  This includes the auto-named cluster node.  If `restoreNodeTimeout` > 0, the cluster goes into an `offline` state until a successful connection can be established.
```
dougwilson: I was confused at first when I read this, until I realized this is about PoolCluster, not with the Pool itself. Can you either move this up into the pool cluster documentation, or move this down to a new paragraph and clarify that it is with PoolCluster?

Issue #2001, opened 2018-04-19
ER_NOT_SUPPORTED_AUTH_MODE: Client does not support authentication protocol requested by server; consider upgrading MariaDB client


MythicalSora:
I'm trying to connect my Discord bot to my server's MySQL database, but everytime I try to run the bot I get the following error:

`let conn = mysql.createConnection({
  insecureAuth : true,
  host : config.mysql.host,
  user : config.mysql.user,
  password : config.mysql.password,
  database : config.mysql.database
});

conn.connect();`

config.json:
`{
    "token":"bottoken",
    "prefix":"m!",
    "mysql": {
        "host":"localhost",
        "user":"root",
        "password":"apassword!",
        "database":"aphi"
    }
}`

MythicalSora:
Where can I see which auth method my server uses ? (Sorry, I'm not that smart when it comes to MySQL & servers)

dougwilson:
No problem. I'm not at a computer right now, but I think it is one of the columns in the mysql.user system table row for the user you are trying to authenticate as.

MythicalSora:
It says my auth method is unix_socket
![image](https://user-images.githubusercontent.com/26851843/39035757-73bf9466-447b-11e8-93ed-f128045fa256.png)

sidorares:
duplicate of #1507

sidorares:
well, kind of, need to confirm if `auth_socket` in mysql is same as `unix_socket` in mariadb. I think both rely on [SO_PEERCRED](http://welz.org.za/notes/on-peer-cred.html)

tulu1688:
Me also. Tried to run in window 10, and mysql server in a docker container. Applied the fix on 1507 but didn't work.

Abhay64:
try the older version of mysql like 5.6.40.1, it uses SHA256_password auth where new version 8.0. uses sha2_password auth which is more secure and throw this authentication protocol error.
[https://dev.mysql.com/get/Downloads/MySQLInstaller/mysql-installer-web-community-5.6.40.1.msi](Mysql installer 5.6.40)

MohammadJozaa7:
I have the same problem with MySQL and I solve by using XAMPP to connect with MySQL and stop the services in windows for MySQL (control panel - Administrative Tools - Services), and in the folder db.js (that responsible for the database ) I make the password empty (here you can see:)

const mysql = require('mysql');
const connection = mysql.createConnection({
  host: 'localhost',
  user: 'root',
  password: ''
});

[...]

PR #2225, opened 2019-05-29
Add support for 'mysql_clear_password'

nmggithub:
This addition is two-fold. First of all, it allows user-set flags to enable flags that were disabled by default. This allows users to enable the PLUGIN_AUTH flag. Second and finally, it adds a case auth handler for the 'mysql_clear_password' auth plugin. This case simply returns the password, in plain text, as a buffer. These  both work in conjunction to allow for users to connect to databases that request the 'mysql_clear_password' plugin. ~~This solves #2001~~ This enables support for MariaDB servers using the [PAM Authentication Plugin](https://mariadb.com/kb/en/library/authentication-plugin-pam), the user just has to add `flags: ['+PLUGIN_AUTH']` to their options object. I could have changed the default for this flag, but I am unaware of the consequences of enabling it by default, so I left it up to the user to enable it. This is tested and works on my machine for a MariaDB database provided by my university for my database course.

### DIFF
--- a/lib/ConnectionConfig.js
+++ b/lib/ConnectionConfig.js
@@ -66,7 +66,7 @@ ConnectionConfig.mergeFlags = function mergeFlags(defaultFlags, userFlags) {
 
   // Merge the new flags
   for (var flag in newFlags) {
-    if (allFlags[flag] !== false) {
+    if (allFlags[flag] !== newFlags[flag]) {
       allFlags[flag] = newFlags[flag];
     }
   }
--- a/lib/protocol/Auth.js
+++ b/lib/protocol/Auth.js
@@ -8,6 +8,8 @@ function auth(name, data, options) {
   switch (name) {
     case 'mysql_native_password':
       return Auth.token(options.password, data.slice(0, 20));
+    case 'mysql_clear_password':
+      return Buffer.from(options.password);
     default:
       return undefined;
   }
--- a/test/integration/connection/test-clear-auth-bad.js
+++ b/test/integration/connection/test-clear-auth-bad.js
@@ -1,6 +1,0 @@
-var assert = require('assert');
-var common = require('../../common');
-
-common.getTestConnection({ flags: ['-PLUGIN_AUTH'] }, function (err) {
-  assert.ok(err, 'got error');
-});
--- a/test/integration/connection/test-clear-auth-bad.js
+++ b/test/integration/connection/test-clear-auth-bad.js
@@ -1,0 +1,6 @@
+var assert = require('assert');
+var common = require('../../common');
+
+common.getTestConnection({ flags: ['-PLUGIN_AUTH'] }, function (err) {
+  assert.ok(err, 'got error');
+});
--- a/test/integration/connection/test-clear-auth.js
+++ b/test/integration/connection/test-clear-auth.js
@@ -1,0 +1,7 @@
+var assert = require('assert');
+var common = require('../../common');
+
+common.getTestConnection({ flags: ['+PLUGIN_AUTH'] }, function (err, connection) {
+  assert.ifError(err, 'got error');
+  connection.destroy();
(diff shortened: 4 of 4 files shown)


nmggithub:
@dougwilson Ok, I would like to help with the tests, but I don't know how to start. I've tested this with my school's database, but I obviously would not want to included my credentials for it in a test. What would we use to test it?

dougwilson:
Hi @nmggithub ! So this module contains an automated test suite. The overview of what to include in a contribution can be found in our contributing section of the README: https://github.com/mysqljs/mysql#contributing

Ideally you could add a new integration test to test against a database that is set up with a clear password auth pugin, but you can also make a unit test that uses the fake server and you'll need to implement how the server does clear auth if you go that route, though.

dougwilson:
Also @nmggithub can you clarify how this solves #2001 ? The user who opened the issue is showing they are using `auth_socket` auth, but this seems to implement a different type of auth from the issue #2001 .

nmggithub:
@dougwilson Apologies for the late reply, as I have been busy. Looking closer, it does seem I was mistaken is saying this solves #2001 , but only inasmuch as it pertains to the `unix_socker` auth. What this is doing is adding support for [MariaDB PAM authentication](https://mariadb.com/kb/en/library/authentication-plugin-pam/#mysql_clear_password). However, the same exact error is given in the case of trying to login to a MariaDB server using PAM authentication, which makes me think my fix is, at least in part, a fix for #2001 . I can play around with a MariaDB install to see if I can replicate the error using `unix_socket` and will update the PR with any additional code that may need to be added.

nmggithub:
@dougwilson Ok I found solution for #2001 (but seeing as I am locked from commenting, I didn't know where to put this, so I'm putting it here). The error in that issue is due to the fact that the auth method is `unix_socket`. MariaDB (and I'm assuming every other flavor of MySQL), [doesn't allow remote connections to users with the `unix_socket` auth method](https://mariadb.com/kb/en/library/authentication-plugin-unix-socket/). This can be seen by enabling the `PLUGIN_AUTH` flag, and seeing the new rejection messages, which see the user trying to login as . So the user that started the issue would have to run an `ALTER USER` query from the machine that is running the server. Then that will allow remote connections to the host.

TL;DR: This pull request does NOT fix #2001 as that is a server-side issue. This pull request DOES add support for MariaDB databases that use the PAM authentication plugin, though.

nmggithub:
@dougwilson tests have been added. Apologies for the multiple commits. I was having issues with my git client. Anyway, I spun up a DigitalOcean droplet, and setup a server there and tested against it. It works, but I don't want to send the credentials publicly here. Is there a way I can privately send them to you / show how to run the tests?

nmggithub:
Deleted bad test.

dougwilson:
Hi @nmggithub sorry I have been away over the weekend. So I don't need your credentials to your own test servers :) The idea for the tests is that they will run on the CI infrastructure here and that they would test the supported features using the CI servers provided by the Travis CI platform. The idea is that if you want to make an integration test (based on the files checked in) then you'll want to probably edit the `.travis.yml` file in this project to create a user which uses the given plugin in which the underlying integration test will then test again.

I hope that makes sense. Please let me know if you need any help. I'll be back online again on Monday :) !

nmggithub:
@dougwilson The issue is the plugin is actually serverside. If a user is authenticated via PAM, as previously explained, the server will ask via that plugin. So I cannot make such a user on the client side.

dougwilson:
Hi @nmggithub that is unfortinate. I will close this PR for now and you can always re-open if you can get some automated tests created. The tests are run on the CI server along with the server-side so it should be possible to do. Perhaps you could even open an issue about this feature to describe it and someone else (perhaps me when I get some time) can help implement the support in this module along with the proper tests?

Issue #2253, opened 2019-08-01
Support non-enumerable `sql` and `values` in query object


blakeembrey:
This issue come up in https://github.com/blakeembrey/sql-template-tag/issues/2, based on the code in https://github.com/mysqljs/mysql/blob/master/lib/Connection.js#L39-L41 the `query` object copies properties verbatim from the input object. However, this is an issue for anyone wanting to have computed properties that are not enumerable (any `get` is not enumerable by default).

Happy if this will be denied, it's workaround-able, but it would be nicer to copy just the properties you need (e.g. `['sql', 'values', 'timeout']`).

PR #2254, opened 2019-08-01
Copy properties from `sql` object explicitly

blakeembrey:
Closes https://github.com/mysqljs/mysql/issues/2253.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -12,6 +12,7 @@ you spot any mistakes.
 * Fix early detection of bad callback to `connection.query`
 * Support Node.js 12.x #2211
 * Support Node.js 13.x
+* Support non-enumerable properties in object argument to `connection.query` #2253
 * Update `bignumber.js` to 9.0.0
 * Update `readable-stream` to 2.3.7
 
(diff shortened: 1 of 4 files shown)


dougwilson:
Do you think using Object.create would be sufficient but still keep from maintaining a list at that location? Adding a test that represents the issue would enable me to answer my own question vs defer to you, ps :)

blakeembrey:
@dougwilson Yes, pretty sure, assuming no one is iterating over the `Object.keys` elsewhere within the project. Where should I add a test case? I kind of hoped the existing tests of cover this, since I'm not changing functionality.

dougwilson:
> Where should I add a test case?

So the test files are in the `test/` directory. You'll probably wanted to add them under `unit` since these are likely unit tests.

> I kind of hoped the existing tests of cover this, since I'm not changing functionality.

Well, if there was an existing test, then I guess this PR is unneeded? Not sure why the test suite would be passing without your change here if there is already a test in this module's suite?

blakeembrey:
> Well, if there was an existing test, then I guess this PR is unneeded? Not sure why the test suite would be passing without your change here if there is already a test in this module's suite?

Derp, right - let me add this.

blakeembrey:
Added a test that I believe should work given node 0.6's feature set.

blakeembrey:
@dougwilson Bump.

dougwilson:
I plan to have a new minor out next week and will get this in there.

PR #2255, opened 2019-08-03
Add missing reflinks to Changes.md

JaanJah:
I noticed that `Changes.md` had reflinks missing from the tags, so this pull request adds them.

Following this reflinks format: https://keepachangelog.com/en/1.0.0/

PS: I noticed that some tags were missing from the repo, but which are in `Changes.md`, so instead of pretending those tags exist, I just didn't add reflinks for those.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -4,16 +4,16 @@ This file is a manually maintained list of changes for each release. Feel free
 to add your changes here when sending pull requests. Also send corrections if
 you spot any mistakes.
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Thank you for your effort, but they weren't actually missing. The ones which have them are because they don't have any notes written in the change log file, so a link to the changes was quickly added. If anything is missing, it's just what the changelog was for those older version that have a link instead.

PR #2305, opened 2020-02-25
Better queryFormat function

barrar:


dougwilson:
Thank you, but this is just an example made to be simple to understand. The idea is that users can write whatever they like for their use case :+1:

PR #2310, opened 2020-02-28
Update documentation around custom formatters.

IrishAdo:
The example worked for me but since 99.9% of our existing project was using ? placeholders I tidied the Function example up to first check if the SQL statement used ? and then customised for that before checking if it should be named parameters.  Also we have one SQL that seems to be using := assignment, I have added a condition for it to not effect that.


dougwilson:
Thank you, but this is just an example made to be simple to understand. The idea is that users can write whatever they like for their use case and the example is simply to demonstrate how to use the API👍

PR #2327, opened 2020-03-28
Master nmggithub

ZJONSSON:
Thanks @nmggithub for the mysql_clear_password branch. This PR is a continuation of https://github.com/mysqljs/mysql/pull/2225 with some unit tests added.     Also I'm rejecting mysql_clear_password unless it's on a secure socket, which seems to be conventional among database clients.   Maybe we should allow an override to that condition, or just drop the check altogether?

@dougwilson I didn't find a simple way to get access to whether the socket is secure or not in a generic fashion. I implemented it in a relatively hacky way here (setting `this._tls = true`) for the handshake, but changeUser might still be missing access to whether the socket is secure or not.


lib/ConnectionConfig.js
```diff
@@ -69,7 +69,7 @@ ConnectionConfig.mergeFlags = function mergeFlags(defaultFlags, userFlags) {
 
   // Merge the new flags
   for (var flag in newFlags) {
-    if (allFlags[flag] !== false) {
+    if (allFlags[flag] !== newFlags[flag]) {
```
dougwilson: Is this really needed? The issue here is that this allows users to now enable flags this module disables because the flag is not actually supported by the module. If the flag is now supported by the module, you can just remove the blacklist of that flag instead of changing this part.

test/FakeServer.js
```diff
@@ -95,7 +95,7 @@ FakeConnection.prototype.handshake = function(options) {
   var packetOptions = common.extend({
     scrambleBuff1       : Buffer.from('1020304050607080', 'hex'),
     scrambleBuff2       : Buffer.from('0102030405060708090A0B0C', 'hex'),
-    serverCapabilities1 : 512, // only 1 flag, PROTOCOL_41
+    serverCapabilities1 : 512 | 1 << 11, // only 2 flags, PROTOCOL_41 and SSL
```
dougwilson: We don't want to enable this for all the tests; if you have a specific test that needs it, that is why this part of the code will allow it to be overwritten by the given handshake options.

lib/protocol/Auth.js
```diff
@@ -2,12 +2,18 @@ var Buffer = require('safe-buffer').Buffer;
...
-function auth(name, data, options) {
+function auth(name, data, options, isSecure) {
   options = options || {};
 
   switch (name) {
     case 'mysql_native_password':
       return Auth.token(options.password, data.slice(0, 20));
+    case 'mysql_clear_password':
+      if (!isSecure) {
+        throw new Error('Authentication method mysql_clear_password not supported on insecure connections');
```
dougwilson: I get where this is coming from, but is probably going to be limiting to users in the long run. What about just adding an option the user can use to enable / disable this? I'm thinking along the lines of our `insecureAuth` option (perhaps `clearAuth`?) and how the `mysql` command line program requires a flag to enable it (https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_enable-cleartext-plugin).

test/FakeServer.js
ZJONSSON: excellent - will update

lib/ConnectionConfig.js
ZJONSSON: This is from the previous PR -  @nmggithub?

lib/protocol/Auth.js
ZJONSSON: It looks like the `--enable-cleartext-plugin` is just to enable the plugin itself.   Are you thinking the `insecureAuth` would allow clear_password even if we are not on a tls socket?

ZJONSSON:
Thanks @dougwilson - just to clarify my interest in this - we are doing a proxy mysql server and want to handle the auth using bcrypt, as credentials are shared with other services.   The only realistic way to use bcrypt is to required clear_password from a secure channel.   Let me take another spin at this, really appreciate all the suggestions you have.

lib/protocol/Auth.js
dougwilson: I'm thinking to have it behave inline with the `mysql` command line program: add an option (like perhaps `clearAuth`?) that that user must enable to use the clear password auth. That would be the only requirement -- the user can choose to use that over SSL or not as their own choice, just like in the `mysql` command line program.

ZJONSSON: I understand.   Looking at the python module, if you specify `auth_plugin='mysql_clear_password'` then cleartext is used whether SSL is on or not.  So maybe that would be the expected behavior.

I still think there is a big difference between saying:  I want to use cleartext only inside secure channel, vs I want to use cleartext no matter what.   For example, some of the GUI Sql clients I tested would bail on cleartext if SSL is not turned on.

Maybe the right way is to use `clearAuth = true` to enable it for secure channel, and `insecureAuth` to allow clear to be in an unencrypted socket?

ZJONSSON:
@jbreckman is it still LGTM?

PR #2341, opened 2020-04-24
Added isConnected method to Connection.js

anaskhanyusfzai:
I have added isConnected() method to Connection.js . Now its easy to check whether you are connected to database or not.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -116,6 +116,15 @@ Connection.prototype.connect = function connect(options, callback) {
   this._protocol.handshake(options, wrapCallbackInDomain(this, callback));
 };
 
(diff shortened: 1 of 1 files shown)


anaskhanyusfzai:
I have worked on diverse tools for development (PHP,Java,C#) and isConnected() method is a common perception to be used for check whether the connection is still connected. I was working on some other project today and found it missing so i added

dougwilson:
Thank you, that is very appreciated. I appreciate that other tools have differing APIs, and acknowledge that each project can have it's own opinion as to what type of API it should present to the user.

anaskhanyusfzai:
Sure , thank you for your time

PR #2359, opened 2020-06-10
Updating link

richardstevens:
Would it be ok if we could update the link to Holidayextras please? We are moving over to .com and https :)


richardstevens:
This is now moved over and live, is it ok to merge this please?

dougwilson (approved):
sorry, apparently i didn't assign it to myself to remember. merging now 🙏

PR #2383, opened 2020-08-10
Fix/error suppress

imanshugupta:
I have a minor change in mysql library which prevents app crashes while using it with nodejs.


dougwilson:
Hi @imanshugupta this is a branch I created that is not ready to be merged, as indicated by the `fixup!` commits still in there and the failing CI. Generally you should only create a pull request from your own branch rather than from others', as you may not always be aware of the state of the code and liekly won't have any permissions to push changes as part of the review process.

PR #2384, opened 2020-08-10
Fix/error suppress

imanshugupta:
I have a minor change in mysql library which prevents app crashes while using it with nodejs.
`Connection.prototype._handleProtocolError = function(err) {
  this.state = 'protocol_error';
  try {
    this.emit('error', err);
  } catch(err) { }
};`

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -31,6 +31,7 @@ you spot any mistakes.
 * Fix `connection.threadId` missing on handshake failure
 * Fix duplicate packet name in debug output
 * Fix no password support for old password protocol
+* Fix suppression of fatal protocol server errors #2021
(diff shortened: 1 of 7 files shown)


dougwilson:
Hi @imanshugupta this is a branch I created that is not ready to be merged, as indicated by the fixup! commits still in there and the failing CI. Generally you should only create a pull request from your own branch rather than from others', as you may not always be aware of the state of the code and liekly won't have any permissions to push changes as part of the review process.

I see this is the second time you've opened a pull request off the same branch. You may be struggling to use GitHub, perhaps. I'm not sure what you're trying to do, but you can find help on how to use GitHub at https://docs.github.com/

PR #2385, opened 2020-08-10
App crashing fix due to protocol error

imanshugupta:
protocol error emits error event which is not properly handleled which causes node app crash.
Its a preventation of app crash


dougwilson:
Closing due to no follow up.

Issue #2350, opened 2020-05-03
Error: UNKNOWN_CODE_PLEASE_REPORT: Connection was killed


Apollon77:
https://sentry.iobroker.net/share/issue/bfbf00157f54442ea5e04695ae4cf661/

```
Error: UNKNOWN_CODE_PLEASE_REPORT: Connection was killed
  File "/opt/iobroker/node_modules/mysql/lib/protocol/sequences/Sequence.js", line 47, col 14, in Quit.Sequence._packetToError
    var err  = new Error(code + ': ' + packet.message);
  File "/opt/iobroker/node_modules/mysql/lib/protocol/sequences/Sequence.js", line 96, col 17, in Quit.Sequence.ErrorPacket
    this.end(this._packetToError(packet));
  File "/opt/iobroker/node_modules/mysql/lib/protocol/Protocol.js", line 291, col 23, in Protocol._parsePacket
    sequence[packetName](packet);
  File "/opt/iobroker/node_modules/mysql/lib/protocol/Parser.js", line 433, col 10, in Parser._parsePacket
    this._onPacket(packetHeader);
  File "/opt/iobroker/node_modules/mysql/lib/protocol/Parser.js", line 43, col 10, in Parser.write
    this._parsePacket(packetHeader);
  File "/opt/iobroker/node_modules/mysql/lib/protocol/Protocol.js", line 38, col 16, in Protocol.write
    this._parser.write(buffer);
  File "/opt/iobroker/node_modules/mysql/lib/Connection.js", line 88, col 28, in Socket.<anonymous>
    connection._protocol.write(data);
  File "/opt/iobroker/node_modules/mysql/lib/Connection.js", line 526, col 10, in Socket.<anonymous>
    fn.apply(this, arguments);
  File "events.js", line 198, col 13, in Socket.emit
  File "domain.js", line 448, col 20, in Socket.EventEmitter.emit
  File "    --------------------", in null.<anonymous>
  File "/opt/iobroker/node_modules/mysql/lib/protocol/Protocol.js", line 144, col 48, in Protocol._enqueue
    sequence._callSite = sequence._callSite || new Error();
  File "/opt/iobroker/node_modules/mysql/lib/protocol/Protocol.js", line 91, col 23, in Protocol.quit
    var sequence = this._enqueue(new Sequences.Quit(options, callback));
  File "/opt/iobroker/node_modules/mysql/lib/Connection.js", line 239, col 18, in Connection.end
[...]

Issue #1982, opened 2018-04-02
UNKNOWN_CODE_PLEASE_REPORT - Code 4025


apopescu-im:
Got this from a failing constraint:

```
--> ComQueryPacket
ComQueryPacket {
  command: 3,
  sql: 'insert into `roles` (`bu_id`, `comment`, `community_id`, `is_admin`, `is_master`, `user_id`) values (1, \'\', 1, false, false, 11)' }

<-- ErrorPacket
ErrorPacket {
  fieldCount: 255,
  errno: 4025,
  sqlStateMarker: '#',
  sqlState: '23000',
  message: 'CONSTRAINT `exactly_one_fk_id` failed for `insights-system`.`roles`' }
```

The failing constraint was defined as:

```
ALTER TABLE roles ADD CONSTRAINT exactly_one_fk_id CHECK (community_id IS NULL XOR bu_id IS NULL)
```

Server version is MariaDB 10.2.12

kai-koch:
This is the correct behaviour for the defined CONSTRAINT and the given example.

Both ```community_id``` AND ```bu_id``` have a value, so the the XOR-Constraint is not meet.
XOR returns false and an error is thrown, which should be caught  and dealt with in your code.

XOR-truth-table
Input | Output
--|--
0, 0 | false
0 , 1 | true
1 , 0 | true
1 , 1 | false

check https://en.wikipedia.org/wiki/Exclusive_or for details of the working of XOR

Closed: no mysqljs error

apopescu-im:
@kai-koch Yes, I'm aware how XOR works.  The given query is intended to fail, as we were testing that the constraint is in place and properly set up.  The reason I posted this is the fact that the error is called "UNKNOWN_CODE_PLEASE_REPORT" so it sounds like a code that needs to have a name defined for it.

Here's the error info:

```
Error handler { Error: UNKNOWN_CODE_PLEASE_REPORT: CONSTRAINT `exactly_one_fk_id` failed for `insights-system`.`roles`
    at Query.Sequence._packetToError (/proj/r2/insights/admin-api/node_modules/mysql/lib/protocol/sequences/Sequence.js:52:14)
    at Query.ErrorPacket (/proj/r2/insights/admin-api/node_modules/mysql/lib/protocol/sequences/Query.js:77:18)
    at Protocol._parsePacket (/proj/r2/insights/admin-api/node_modules/mysql/lib/protocol/Protocol.js:279:23)
[...]

PR #2386, opened 2020-08-10
Add error codes from MySQL 8

### LINKED ISSUE #2325: UNKNOWN_CODE_PLEASE_REPORT - 3643
`Geometry SRID (0) does not match column SRID`.
The error should be named `ER_WRONG_SRID_FOR_COLUMN` (=3643)

ARitz-Cracker:
Fixes #2325. #2350

This PR makes the `generate-error-constants.js` tool able to get all the error codes from the MySQL 8 sources, as the txt file which defines all of them was in a different location compared to MySQL 5.

I've also included the output from the tool.

However, there appears to be one issue. The MySQL 8 sources marks a lot of the MySQL 5 error codes as "obsolete". Since MySQL 5 is going to go EOL in Oct 2023, I'm assuming that these "obsolete" codes shouldn't be marked as such, at least until then. In order to solve this, we can do one of the two things:

1. Manually edit `errors.js` to include the non-obsolete definitions for the MySQL 5 error codes.
2. Allow for `generate-error-constants.js` to be passed 2 directories, one pointing the the MySQL 5 sources, and the other to the MySQL 8 sources. If it detects a MySQL 8 definition which starts with `OBSOLETE_`, it'll fall back to the MySQL 5 definition.

Please let me know what's preferred so I can make the adjustments.


dougwilson:
Thank so much @ARitz-Cracker ! Yea, we definately need some solution to the 5 vs 8 issue, as we cannot just drop this change in as-is with changing all those code to include the word obsolete -- as code that uses this library relies on the code names there for detection of what type of error occurred vs the number, so it would be at least a semver major change for this module (not even considering if it makes sense to lead them with obsolete).

Perhaps, if all the codes are included in 5 sources still, would just having the tool simply remove the `OBSOLETE_` prefix be sufficient?

As for your linked issues, I see you say this fixes those two bug reports. Can you list here which code number fixes #2325 and which code number fixes #2350 ? We don't want to close those out without being sure this fixes those issues, and the way you have the description written GitHub will automatically close them once this is merged, so want to be sure it makes sense.

ARitz-Cracker:
Hello, @dougwilson!
This PR does include `ER_WRONG_SRID_FOR_COLUMN` which #2325 refers to. I'll remove the reference to #2350 as I can't find the string "Connection was killed" on its own anywhere in the MySQL 8 sources.

As for making the tool simply removing the `OBSOLETE_` prefix, I don't think that's a good idea. The most obvious reason being that some 5.x codes are defined as "OBSOLETE_ER_UNUSED" in 8.x. [like here](https://github.com/mysqljs/mysql/pull/2386/files#diff-8ee6343a449442cdd76580c85dede8f0R2118)

Going down this path, I think making the tool read both the 5.x sources and the 8.x sources is the best solution.

Also, the CI is failing due to linting errors, I'll make sure to run eslint this time. 😆

ARitz-Cracker:
By the way, you may notice that this PR still replaces some of the generated output with "unused" or "obsolete", this must have been done sometime between 5.7.29 and 5.7.31. So there are still some definitions which both versions mark as obsolete. Should I make the script remove those last remaining obsolete and unused definitions from the list?

ARitz-Cracker:
@dougwilson one more thing! Since I'm going through the trouble of modifying the error constants generator anyway, I can further extend the tool to read the MariaDB sources as well, and thus solve #1982! the MariaDB 10 sources have them defined in a similar manner as MySQL 5 with some minor tweaks. So it definitely won't be difficult to do.

dougwilson:
That would be neat. I think I tried that and ran into issues. There may be a branch in this repo with that work already if you want to take a look.

But regardless, please make sure to do the MariaDB changes as a separate PR from this one.

I have been at work since I first commented and will take a look at the new changes of this PR after work.

ARitz-Cracker:
Sounds great! I'll create a new branch which extends onto this one. Also, just made one final change to make sure that this tool will always prioritize MySQL 5 definitions over MySQL 8 so that there are no breaking changes.

ARitz-Cracker:
I've made another branch [here](https://github.com/ARitz-Cracker/mysql/tree/mariadb-err-consts), which contains the MariaDB codes and the MySQL 8 codes. I'll submit a PR for that one once this one is approved.

lib/protocol/constants/errors.js
```diff
@@ -269,8 +333,8 @@ exports.ER_NO_SUCH_TABLE
 exports.ER_NONEXISTING_TABLE_GRANT                                                       = 1147;
 exports.ER_NOT_ALLOWED_COMMAND                                                           = 1148;
 exports.ER_SYNTAX_ERROR                                                                  = 1149;
-exports.ER_DELAYED_CANT_CHANGE_LOCK                                                      = 1150;
-exports.ER_TOO_MANY_DELAYED_THREADS                                                      = 1151;
+exports.OBSOLETE_ER_UNUSED1                                                              = 1150;
```
dougwilson, line 336: There still seems to be something wrong here, as it seems these are still being changed to say obsolete.

ARitz-Cracker: That's what it says in 5.7.31! I downloaded 5.7.29 and it's also marked as "UNUSED1". So those are probably from an even older version. Checking the blame here, I see you purposefully omitted these kinds of re-names in the past.

dougwilson: hmmm, strange. I don't recall manually omitting changes, but it's certainly possible :)

dougwilson: As a check I just re-downloaded the MySQL 5.7.29 source code and ran the current tool against that source tree and there is no changes produced to this file, so something more is going on here 🤔

ARitz-Cracker: Ah, so this the tool is supposed ignore previously defined codes that are marked as "unused". See the `keepUnusedCodes`, it checks for `ER_UNUSED[0-9]*`, and falls back to what the file had previously. In `fixupCode`, all the `OBSOLETE` markers are supposed to be removed, but it only checks if the string starts with `ER_OBSOLETE_`, not just `OBSOLETE_`. That would be fine since MySQL 5 defines unused codes as `ER_UNUSED[0-9]*`. _However_ my first commit ran the tool with _only_ MySQL 8. MySQL 8 started using `OBSOLETE_ER_UNUSED[0-9]*` instead of `ER_UNUSED[0-9]*`. Subsequent commits perfecting the tool were then using a version of `errors.js` with the older codes already removed.

The solution?

Re-run the new tool with an completely un-altered version of the `errors.js` file from master.

ARitz-Cracker:
I've just re-written history. I removed the changes to `errors.js` during my 3 wip commits, and added a final commit where I ran the tool. Things look like they work as expected now!

lib/protocol/constants/errors.js
```diff
@@ -1239,6 +1303,494 @@ exports.ER_XA_REPLICATION_FILTERS
...
+exports.ER_EXCEEDED_MV_KEYS_SPACE                                                        = 3906;
+exports.ER_FUNCTIONAL_INDEX_DATA_IS_TOO_LONG                                             = 3907;
+exports.ER_WRONG_MVI_VALUE                                                               = 3908;
+exports.ER_WARN_FUNC_INDEX_NOT_APPLICABLE                                                = 3909;
+exports.ER_GRP_RPL_UDF_ERROR                                                             = 3910;
+exports.ER_UPDATE_GTID_PURGED_WITH_GR                                                    = 3911;
+exports.ER_GROUPING_ON_TIMESTAMP_IN_DST                                                  = 3912;
+exports.ER_TABLE_NAME_CAUSES_TOO_LONG_PATH                                               = 3913;
+exports.ER_AUDIT_LOG_INSUFFICIENT_PRIVILEGE                                              = 3914;
+exports.OBSOLETE_ER_AUDIT_LOG_PASSWORD_HAS_BEEN_COPIED                                   = 3915;
```
dougwilson, line 1721: Shouldn't there not still be keys that are starting with `OBSOLETE_` in here?

ARitz-Cracker: Ah, these were marked as obsolete before 8 was released, but didn't exist in 5 so I didn't think it mattered. But! I've made the tool now auto-remove the `OBSOLETE_` prefix if found!

ARitz-Cracker:
Hello! I don't know why the AppVeyor build failed. Should I rebase this PR?

PR #2415, opened 2020-09-30
Update Readme.md

vishal1009:


dougwilson:
The title of the readme matches the npm module name, since this is a npm module.

PR #2442, opened 2020-12-04
Adding current GA version to test suite

rusher:
change statistics test (mariadb doesn't have flush_tables)
change load data infile disable (MariaDB return a specific error 4166 : ER_LOAD_INFILE_CAPABILITY_DISABLED)
ensure load data is enable server side (mysql 8/mariadb 10.5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,10 @@ matrix:
       env: "DOCKER_MYSQL_TYPE=mariadb DOCKER_MYSQL_VERSION=10.2"
     - node_js: *lts
       env: "DOCKER_MYSQL_TYPE=mariadb DOCKER_MYSQL_VERSION=10.3"
+    - node_js: *lts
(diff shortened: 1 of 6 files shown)


rusher:
error in travis is because of a different error (common.fakeServerPort 32893 was already used)
(FakeServer test would probably be better using port 0, then getting resolved port)

grooverdan:
Thanks @rusher for setting me on the right track.

https://github.com/MariaDB/mariadb.org-tools/blob/master/buildbot.mariadb.org/dockerfiles/ecofiles/test-mysqljs.sh#L43-L56 is the changes to integration tests that I needed to do to get it successful on all mariadb versions (CI tests: https://buildbot.mariadb.org/#/builders/123).

As such test-statistics.js needs a `common.minVersion(connection, 10, 5, 0)` and the other tests, test-bad-credentials.js and test-server-timeout-disconnect.js needing version dependent tests.

PR #2480, opened 2021-04-18
Remove queued query

### LINKED ISSUE #2478: Remove queued query
Would it be possible to add a connection method that would remove a queued (waiting) query from the connection queue? It would take a single parameter the query object as returned from `connection.query()` and it would remove the query from the queue if the query is waiting. It should respond with one of three states: query removed, query currently executing (can't be removed), query not found (probably already executed).

A rough pseudocode what it would do (specific implementation details will need to be refined):

```
Connection.prototype.remove = function remove(query) {
  var idx = this._protocol._queue.indexOf(queue)
  if (idx < 0) {
    return false; // Query not found
  }
  if (idx === 0) {
    return false; // Query currently executing
  }
  this._protocol._queue.splice(idx, 1)
  return true;
}
```

We need this feature for implementing timeouts in Knex.js, see this discussion https://github.com/knex/knex/pull/4416#discussion_r608067856. We don't want to hack the driver internals and we would prefer to use a supported feature. I'm willing to create a PR with the implementation if the feature would be accepted.

martinmacko47:
Resolves  #2478


martinmacko47:
I will add tests later, after the general approach in this implementation is decided acceptable.

lib/protocol/Protocol.js
```diff
@@ -336,6 +336,14 @@ Protocol.prototype._dequeue = function(sequence) {
     return;
   }
 
+  if (sequence !== this._queue[0]) {
+    var idx = this._queue.indexOf(sequence);
+    if (idx > 0) {
```
maximelkin, line 341: >= ?

martinmacko47: `idx` will never be `0` as `sequence !== this._queue[0]`. The sequence in `this._queue[0]` is the one that is currently being executed, isn't it?

Or you mean to change it to `idx >= 1`?

lib/protocol/sequences/Query.js
```diff
@@ -24,12 +24,29 @@ function Query(options, callback) {
...
+
+Query.prototype.ended = function() {
+  return this._ended;
+};
+
+Query.prototype.dequeue = function() {
+  if (this._started) {
+    throw new Error('Query already started.');
+  }
+  this.end();
```
martinmacko47, line 47: I'm not entirelry sure calling `this.end()` here won't mess something. I've tracked the code, it seems to behave as expected: it calls `this._callback` with no results, emits `end` which calls `Protocol._dequeue(this)`. But I'm not sure I covered all cases that may happen.

lib/protocol/Protocol.js
maximelkin: My bad, missed previous line.

Btw maybe use separate method for removing sequence from queue? Its seems current method executes next sequence, but im not sure.

martinmacko47: @maximelkin I don't call `Protocol._dequeue(sequence)` directly. It is called from handler of `Sequence.emit('end')` emited from `Query.end()` called in my `Query.dequeue()`. I wanted to use the native query ending mechanism, so the query will call its `_callback` (with nothing) after being cancelled.

But yes, if we think it would be better, we could short circuit the native query ending mechanism, and just remove the query from the queue without calling `Query.end()` and emitting `Sequence.emit('end')`. But then the query `_callback` would never be called. I'm not sure if it would be good or not.

PR #2498, opened 2021-06-30
Password as a function

kyawlinmyat:
Support password as a function both sync and async method.


dougwilson:
Hello, I was confused for a while why it said I closed this PR, but it was apparently because the PR was against `develop` branch instead of `master`. You're welcome to open it again against `master`, but ideally with docs and tests this time :)

PR #2507, opened 2021-07-16
release idle connections after timeout

yqsailor:
Hello, I add `maxIdle` and `idleTimeout` to pool options, making it supports automatically released the idle connections after timeout. I think this feature is very useful, hope you merge and release a new mysql version in the near future. If you have any questions, some feedback would be appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 coverage/
 node_modules/
 npm-debug.log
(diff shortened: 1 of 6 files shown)


yqsailor:
I reference the druid source code: https://github.com/alibaba/druid

PR #2516, opened 2021-08-09
Change UNKNOWN_CODE_PLEASE_REPORT to identify uknown code

glensc:
Changes `UNKNOWN_CODE_PLEASE_REPORT` to include `packet.errno` in it.

This makes it easier to identify that the code is even if the code is not yet supported:

```
(node:1) UnhandledPromiseRejectionWarning: Error: UNKNOWN_CODE_PLEASE_REPORT: Query execution was interrupted, query_timeout exceeded
```


lib/protocol/sequences/Sequence.js
```diff
@@ -43,7 +43,7 @@ Sequence.prototype.hasErrorHandler = function() {
 };
 
 Sequence.prototype._packetToError = function(packet) {
-  var code = ErrorConstants[packet.errno] || 'UNKNOWN_CODE_PLEASE_REPORT';
+  var code = ErrorConstants[packet.errno] || 'UNKNOWN_CODE_' + packet.errno + '_PLEASE_REPORT';
```
glensc, line 46: The other alternative i've considered was:

```suggestion
  var code = ErrorConstants[packet.errno] || 'UNKNOWN_CODE_PLEASE_REPORT: ' + packet.errno;
```

vlasky: I think your alternative is better because any error handling code just has to check for the static string "UNKNOWN_CODE_PLEASE_REPORT".

glensc: @vlasky applied.

Do you have the power to merge and release this?

glensc:
@dougwilson 19aa86d

PR #2564, opened 2023-11-21
Use WHATWG URL parsing, if available, instead of url.parse

### LINKED ISSUE #2563: Updating URL parsing to WHATWG URL
@dougwilson I think we should update connection configuration URL parsing to use the new method as the previous is deprecated and I see some issues open about it (#2561), as well as myself hitting this problem with a more complex URL that `url.parse` is unable to process.

I can do this change right now but if we're going to do this, we should also increase the minimum node version that people can use with this module. What are your thoughts about it?

dresende:
The strategy:

1. Check if `URL` is a function
2. Check if `URL.prototype` is an object

If the checks are true, use `new URL` instead of `url.parse`.

The only 2 changes I know of, are:

1. `URL` has `username` and `password` already parsed, no need to check for `auth`
2. `URL` has `searchParams` instead of `query`

Closes #2563

### DIFF
--- a/lib/ConnectionConfig.js
+++ b/lib/ConnectionConfig.js
@@ -187,11 +187,7 @@ ConnectionConfig.parseUrl = function(url) {
 
   if (typeof url.username == 'string') {
     options.user     = url.username;
-    try {
-      options.password = decodeURIComponent(url.password);
-    } catch (e) {
-      options.password = url.password;
-    }
+    options.password = decodeUriComponent(url.password);
   } else if (url.auth) {
     var auth = url.auth.split(':');
     options.user     = auth.shift();
@@ -224,3 +220,13 @@ ConnectionConfig.parseUrl = function(url) {
 
   return options;
 };
+
+function decodeUriComponent(str) {
+  return str.replace(/\%([a-f0-9]{2})/ig, function (_, hex) {
+    try {
+      return String.fromCharCode(parseInt(hex, 16));
+    } catch (e) {
+      return _;
+    }
+  });
+}
--- a/lib/ConnectionConfig.js
+++ b/lib/ConnectionConfig.js
@@ -177,7 +177,7 @@ ConnectionConfig.parseFlagList = function parseFlagList(flagList) {
 };
 
 ConnectionConfig.parseUrl = function(url) {
-  url = urlParse(url, true);
+  url = (typeof URL == 'function' && typeof URL.prototype == 'object' ? new URL(url) : urlParse(url, true));
 
   var options = {
     host     : url.hostname,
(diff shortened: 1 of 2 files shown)


dougwilson:
Make sure to add a new test case for what you are trying to fix with this. Also curious on the change to the existing test: would this change now break anyone who has a colon in thier password?

dresende:
> Make sure to add a new test case for what you are trying to fix with this. Also curious on the change to the existing test: would this change now break anyone who has a colon in thier password?

That's an interesting question. I did this in kind of a hurry but then figured I could just pass an object and avoid the problem I was having. I'm not in a hurry now, so I'll sure add a test for that. About people using and it breaking, I think I need to change the parser to decode the password when using WHATWG URL as it seems it is always encoded when parsed.

dresende:
```sh
> new URL("foo://user:pa:ss@host")
URL {
  href: 'foo://user:pa%3Ass@host',
  origin: 'null',
  protocol: 'foo:',
  username: 'user',
  password: 'pa%3Ass',
  host: 'host',
  hostname: 'host',
  port: '',
  pathname: '',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
> new URL("foo://user:pa%3Ass@host")
URL {
  href: 'foo://user:pa%3Ass@host',
  origin: 'null',
  protocol: 'foo:',
  username: 'user',
  password: 'pa%3Ass',
  host: 'host',
  hostname: 'host',
  port: '',
  pathname: '',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

It always encodes, whether it's already encoded or not.

dresende:
What was failing for me was that I had a `%` on the password and the URL was not encoded, and because after that character there wasn't necessarily 2 hex characters, `url.parse` fails because it tries to decode and fails.

In the following example, assume password is `pass%` (not encoded).

```sh
> url.parse("foo://user:pass%@host")
Uncaught URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at Url.parse (node:url:368:19)
    at Object.urlParse [as parse] (node:url:156:13)
> new URL("foo://user:pass%@host")
URL {
  href: 'foo://user:pass%@host',
  origin: 'null',
  protocol: 'foo:',
  username: 'user',
  password: 'pass%',
  host: 'host',
  hostname: 'host',
  port: '',
  pathname: '',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

What's failing:

```sh
> decodeURIComponent("pass%")
Uncaught URIError: URI malformed
    at decodeURIComponent (<anonymous>)
```

dougwilson:
But shouldn't you always need to url encodure your password? What if your password is `pa%0foo` ?

dresende:
> But shouldn't you always need to url encodure your password? Whay is your password is `pa%0foo` ?

In that case you should, but according to the spec you don't need to unless it causes ambiguity (like your case). Password must be encoded if it has `%` and 2 hex chars next to it.

dougwilson:
Gotcha. What are we expecting the outcome of `mysql://foo:bar%fzo%25ab@localhost` to be? I know you said that encoding the `%` is not required unless there are two hex chars after it, but this one seems weird when put through the code here, as it seems that `%25` is ending up in the password instead of it being just a `%` character, though `new URL` does parse it now. I think the way `decodeURIComponent` is used may need to be changed to match how the `new URL` is working so they are supporting the same semantics.

dresende:
> I think the way decodeURIComponent is used may need to be changed to match how the new URL is working so they are supporting the same semantics.

I don't know the implementation. Do you have any hints? As a funny note, your example URL is parsed in a different way on Chrome. I thought the implementation was the same.

Screenshot 2023-11-21 at 14 59 37

dresende:
@dougwilson you meant something like this?

```js
function decodeUriComponent(str) {
	return str.replace(/\%([a-f0-9]{2})/ig, (_, hex) => (String.fromCharCode(parseInt(hex, 16))));
}
```
(but converted to run on older node versions)

PR #2569, opened 2024-06-25
Update ssl_profiles.js

whatadewitt:
update Amazon RDS profiles to latest

context: working on a project with legacy codebase. need to update to a newer cert by aug 22, 2024. i can use a fork to handle this, but figure it might save some people some of the headaches i've been dealing with.

### DIFF
--- a/lib/protocol/constants/ssl_profiles.js
+++ b/lib/protocol/constants/ssl_profiles.js
@@ -1,1480 +1,2886 @@
-// Certificates for Amazon RDS
+'use strict';
+
+// CA Certificates for Amazon RDS
+// https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html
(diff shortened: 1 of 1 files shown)


whatadewitt:
If this seems ok, let me know and I can submit with an updated version as well.

sidorares:
@whatadewitt maybe worth switching to https://github.com/mysqljs/aws-ssl-profiles - see some context in https://github.com/sidorares/node-mysql2/issues/2581

dougwilson:
Ya, there isn't a huge reason to update the built in certs here, as the new module splits them out. The separate module is great too bc then updating certs is separate from the version of the mysql module version as wlel.

https://github.com/mysqljs/aws-ssl-profiles?tab=readme-ov-file#mysqljsmysql is the usage example for how to uae the newest certs with thia module if it helps.

PR #2590, opened 2026-09-20
Forward query write failures such as EPIPE to the query callback

### LINKED ISSUE #2021: Write error when executing a too big query
When I try to import a too big list of multi line queries (eg. coming from a sql file).

It will crash:

> { Error: write EPIPE
>     at WriteWrap.afterWrite [as oncomplete] (net.js:835:14) errno: 'EPIPE', code: 'EPIPE', syscall: 'write', fatal: true }

Regards,
Melroy van den Berg

dyk1454683243-sudo:
Fixes #2021

## Problem

Importing or executing a too-large multi-statement / large SQL payload can fail while the query is being written to the socket. Today that `write EPIPE` (or a similar stream write failure) can:

1. Throw out of `socket.write()` and miss the query callback entirely
2. Arrive after the sending sequence has already completed, then bubble as an unhandled `connection` `error` and crash the process even when a query callback was attached

The original report was a large SQL file sent as one query. The server rejected the packet (`max_allowed_packet`) and closed; leftover writes then failed with `{ Error: write EPIPE, fatal: true }`.

A previous attempt to fix this (`74f37f3` on `fix/error-suppress`) also changed packet-number tracking. That was backed out of `master` because it exposed a MariaDB protocol difference in CI. This PR does **not** revive that parser change.

## Fix

Minimal write-path handling only:

- Wrap protocol writes in a safe helper (`try/catch` + `write` callback + `writable` check)
- Deliver in-flight write failures to the current query sequence / callback (`fatal: true`)
- If the write fails after that sequence has already ended, record it as a fatal connection error and close — do **not** emit an unhandled `connection` `error`

Handshake, `ECONNREFUSED`, and other non-write network errors keep their existing semantics.

## Tests

Unit tests mock the socket write path (no huge payload required):

- `test-query-write-epipe.js` — `write()` throws `EPIPE`
- `test-query-write-epipe-event.js` — socket `error` event during send
- `test-query-write-epipe-callback.js` — `write()` callback error
- `test-query-write-epipe-after-error.js` — leftover `EPIPE` after the query callback already ran

## Local test run (Node.js 22.14.0)

- Lint: pass
- New EPIPE unit tests: pass
- Other unit tests: pass except 4 pre-existing SSL fixture failures (`test/fixtures/server.crt` expired 2024-05-11: `certificate has expired`)
- Integration tests: skipped (`cannot connect to MySQL server` — no local mysqld in this environment)


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

